PR #6, opened 2026-06-26
Support custom OpenSpec schemas

nthansen:
## Why

spek hardcoded the four `spec-driven` artifacts (`proposal` / `design` / `tasks` / `specs`) in every layer, so a change authored under any other OpenSpec schema (e.g. `superpowers-bridge` with `brainstorm` / `plan` / `verify` / `retrospective`) had its extra artifacts **silently dropped**.

> **Scope note:** this is the "custom schema" half of the split you asked for in #6. It's grown a bit past a pure split — it also carries the user-selectable ordering we designed on top, and a few correctness/hygiene fixes surfaced while doing it (all listed below). Happy to peel any of it out if you'd rather review a thinner PR. The version bumps / `## 1.4.0` CHANGELOG headers / CRLF restore are removed per your notes; changelog text is at the bottom as release-note input.

## What changed

- **Filesystem-first discovery** — a change's artifacts are discovered from disk (any `*.md` at the change root + the `specs/` delta tree), so every artifact renders regardless of schema, with zero spek changes per new schema.
- **User-selectable ordering** — tabs default to **Last modified** (mtime, newest first) so the artifact you're actively editing surfaces first; a sort control also offers **Schema order** and **A–Z**, persisted globally in `localStorage`.
- **Schema order from the OpenSpec authority** — "Schema order" sources the sequence from the `openspec` CLI (`openspec status --change <slug> --json` → `planningArtifacts` mapped to discovered artifacts), computed **once per change-detail read** and exposed as `ChangeDetail.schemaOrder`. Never on the scan/list/overview hot path. When the CLI is absent or the change is archived it **degrades gracefully** to the default order with a reason-specific message.
- **Non-blocking CLI call** — `readChange` is async and the CLI is spawned via `cross-spawn` with an **argv array (no shell)**, so it can't block the event loop and there's no shell-injection surface (matches the Kotlin `ProcessBuilder` path).
- **Schema badge** per change; **generic `ChangeDetail.artifacts[]` contract** (breaking *internal* API/adapter change); full-text search indexes every markdown artifact.
- Applies across **all four surfaces** (core, web, VS Code, IntelliJ) + the static demo.
- **Hygiene:** removed dead exports this change introduced (kept the `@spek/core` surface to what's actually consumed).

## How the ordering design evolved (the transition to sort-by)

The first cut of this delegated *the* tab ordering to the CLI. Two problems drove a redesign:

1. **CLI-on-the-hot-path was heavy and pinned the wrong artifact first** — ordering by the schema's authoring sequence puts `tasks` last even while you're checking boxes during a run. So the default moved to **filesystem mtime** (newest first): schema-agnostic, no subprocess, the artifact you're touching floats up.
2. **Schema order is still useful** for reading a change in its authored sequence — so instead of picking one, ordering became **user-selectable** (`Last modified` default / `Schema order` / `A–Z`). Schema order keeps the authoritative CLI source but only as an on-demand, non-blocking mode with a visible fallback.

## Verification

- `@spek/core` + web unit tests + Kotlin tests all pass **without** the `openspec` CLI installed (the order provider is injected/faked). Mutation testing on the new/changed core + web modules holds ~91%+ (artifact-sort 95.8%, artifacts 92.5%, schema-order 93.3%).
- Type-check + web / webview / vscode / demo builds green; the bundled VS Code webview and `demo.html` were rebuilt so both carry the sort control.

## Notes

- `action.yml` installs `openspec` best-effort (telemetry off) so static-site builds get authoritative schema order; the build still succeeds via default ordering if that install fails.
- The `plugin.xml` change is **not** a version bump — it's the Changes-Browser feature-description line updated for custom-schema support. (You listed `plugin.xml` under the version bumps; there's no `<version>` change in it.)
- OpenSpec change authored via the workflow, archived to `openspec/changes/archive/2026-07-04-add-artifact-sort-options/` and its delta specs synced into `openspec/specs/`.

### Changelog text (for release notes; not committed)

- Support custom OpenSpec schemas: a change's artifacts are discovered from disk, so any schema's artifacts render as their own tabs.
- Artifact tabs default to last-modified order; a sort control adds Schema order (from the OpenSpec CLI, non-blocking, with graceful fallback) and A–Z.
- Each change shows its schema as a badge; full-text search indexes every markdown artifact.


kewang:
Hi @nthansen — first off, sorry for the slow reply, and thank you for such a thorough contribution! I really appreciate that both pieces were authored through the OpenSpec workflow with their own archived changes, solid test coverage, and honest notes about what was verified where. This is exactly the kind of PR I hoped the workflow would enable.

I'd like to take you up on your offer to split this up. Two asks:

**1. Split into two PRs**

- **PR A — watcher polling fallback** (`f4363b7`): low-risk, well-verified fix. I'd like to review and merge this one first.
- **PR B — custom OpenSpec schemas** (`b4adc35`): great direction, but it's a bigger change (internal `ChangeDetail` API break, spawning the `openspec` CLI at viewer runtime, `action.yml`), so I want to give it a closer review on its own.

Since the two commits are already cleanly separated, this should hopefully be mostly mechanical on your side.

**2. Drop the version bump / CHANGELOG version sections**

Releases here go through my release flow, which writes the CHANGELOG sections and decides the version number at release time — so please remove the `1.3.2 → 1.4.0` bumps (root + vscode `package.json`, `gradle.properties`, `plugin.xml`) and the `## 1.4.0` CHANGELOG headers from the PRs. Your changelog entry text is genuinely useful though — keeping it in each PR description works great as input for the release notes. (You're likely right that this lands as a minor bump; I just want the version/timing decision to stay with the release flow.)

The CRLF line-ending restore is appreciated but please keep it out of the feature commits too — feel free to open it as a tiny separate PR, or I can handle it.

On `action.yml`: the best-effort `openspec` install with graceful fallback seems reasonable to me — let's keep it in PR B and review it there rather than dropping it.

Thanks again for the quality work here — looking forward to landing both!

kewang:
## Code review — custom OpenSpec schemas + artifact ordering

Nice work — the filesystem-first discovery is the right call, the CLI spawn is safely structured (argv array + cross-spawn on TS, arg list + whitelist on Kotlin), the contract migration is clean (no consumer left reading the old `proposal/design/tasks/specs` fields), and demo.html + the vscode webview were rebuilt. All suites pass on my machine: `@spek/core` 76, `@spek/web` 31, IntelliJ Kotlin 20 (ArtifactDiscoveryTest 7 + SchemaOrderTest 13), type-check + all builds green.

Findings ranked by severity. Items marked **reproduced** I ran against the built `@spek/core`; the IntelliJ threading items are from tracing the actual call path.

### 1. `countArtifacts` reads every `spec.md` in full on the changes-list hot path, and crashes the whole list on one unreadable file — confirmed, reproduced
`packages/core/src/artifacts.ts` (`countArtifacts` L51 → `readSpecsTree` L57/L24)

`countArtifacts` calls `readSpecsTree`, which `readFileSync`s every `specs/**/spec.md`, only to test `length > 0` — contradicting its own doc ("不讀取內容"). It runs once per change via `scanChangeDir`, i.e. over every active + archived change for `/changes` and `/overview` (× worktrees under aggregation). Worse, any single unreadable `spec.md` throws out of `countArtifacts` and aborts enumeration of the entire list.

```
countArtifacts(dir) threw: EACCES open spec.md
→ proves it readFileSync's spec.md (a pure existsSync check wouldn't throw)
→ one unreadable spec.md breaks the whole changes list
```

**Fix:** existence-only short-circuit (a `hasSpecsTree` that returns on the first `spec.md`); `specsMtime` already shows the `existsSync`-only pattern.

### 2. `schema-order` module cache is never invalidated or bounded — confirmed
`packages/core/src/schema-order.ts` (L83) and IntelliJ `SchemaOrder.kt` (L26)

The process-global `cache` is only ever get/set. Consequences in a long-running web server / VS Code host: (a) if `openspec` isn't on PATH at the first call, `null` is cached **forever** — installing it later never helps until restart; (b) if a change's artifact order later changes, the stale order is served indefinitely (the comment says "同一次掃描內" but the Map is global); (c) unbounded growth, one entry per `(repoRoot, slug)` incl. failures. Live-reload refreshes everything else but not this. The Kotlin side defines `clearCache()` but **nothing in the project calls it**, so it has the same problem. Suggest a TTL and/or clearing the key on the file-watch event.

### 3. IntelliJ blocks the built-in server handler thread on the CLI, with a `readText()`-before-`waitFor()` deadlock risk — confirmed (path traced)
`ChangeReader.kt` (L28) + `SchemaOrder.kt` (L107)

TS deliberately made `readChange` async and the provider a non-blocking spawn with a real timeout that kills regardless. The Kotlin port runs it synchronously: `SpekHttpRequestHandler.process` → `handleChangeDetail` → `ChangeReader.read` → `SchemaOrder.cli.order` (`ProcessBuilder` + `waitFor(10s)`) all on the shared built-in-server (Netty, :63342) thread. Opening any active change ties that thread up for the CLI duration. And `SchemaOrder.kt:107` does `proc.inputStream.bufferedReader().readText()` **before** `waitFor`, with `redirectErrorStream(false)` and stderr never drained — if `openspec` writes > the ~64KB stderr pipe buffer while keeping stdout open, the child blocks on stderr, `readText()` blocks forever, and the 10s `waitFor`/`destroyForcibly` are never reached. **Fix:** offload to a pooled thread; use `redirectErrorStream(true)` (or drain stderr) and only read stdout after the timeout-bounded wait.

### 4. Root `specs.md` collides with the `specs/` delta tree on Map key `"specs"` → silent data loss + count/discover mismatch — confirmed, reproduced
`packages/core/src/artifacts.ts` (L97)

A change with both a root `specs.md` and a non-empty `specs/`: the root loop sets `built.set("specs", markdown)`, then the specs-tree overwrites it with the specs-kind artifact. The `specs.md` content is dropped, and `countArtifacts` (counts `specs.md` as a `.md` **and** +1 for the tree) disagrees with `discoverArtifacts`.

```
files on disk: proposal.md, specs.md(root), specs/foo/spec.md
discoverArtifacts → [proposal(markdown), specs(specs)]   // specs.md content GONE
root specs.md content still present in any artifact: no ❌
countArtifacts = 3   vs   discoverArtifacts.length = 2   ❌ mismatch
```

Niche (a custom schema would have to name a planning artifact literally `specs`), but it's exactly the "artifact silently dropped" failure this PR sets out to fix. Reserve the `specs` id for the tree, or detect the collision and rename.

### 5. Fresh checkout doesn't fall back to narrative order — plausible
`packages/core/src/artifacts.ts` (L101-111)

The `DEFAULT_ORDER` tiebreak only fires when mtimes are **exactly equal**, but `git clone`/`checkout` writes files sequentially with distinct sub-ms mtimes (ext4/APFS), so on a never-edited change the default ("Last modified") tabs come out ~reverse-checkout order (tasks/specs first), not the proposal→design→… the comment promises. Either fix the comment, or seed default order from the schema for active changes.

### Minor / cleanup
- **Duplicated constant/logic:** `DEFAULT_ORDER` is defined in both `artifacts.ts` and `web/src/utils/artifact-sort.ts`; the "root `*.md`, skip dotfiles, `.md` only" predicate is open-coded in 4 places (`artifacts.ts` count + discover, `web/.../openspec.ts` search, `vscode/src/handler.ts` search). Changing the rule needs all 4 in sync — a `@spek/core` `listChangeMarkdownFiles()` + exported `DEFAULT_ORDER` would prevent the "tab shows but isn't searchable" drift this PR fixes.
- **Stale doc-comment:** `types.ts` (~L72) still says `ChangeArtifact` ordering "由 openspec CLI 提供", but after `order-artifacts-by-recency` the default is mtime; the CLI feeds only the optional `schemaOrder`.
- **Misleading fallback message:** `ChangeDetail.tsx` (~L224) hardcodes "OpenSpec CLI not available" for an active change, but `schemaOrder` can be empty for other reasons (no `planningArtifacts`, unmapped `outputPath`, timeout, non-zero exit) — the CLI may be present and working. The `*` tooltip ("Schema order unavailable") is accurate; the banner over-claims the cause.
- **Kotlin/TS divergences:** `SchemaOrder.cache` is an unsynchronized `HashMap` (Netty concurrency → `ConcurrentModificationException`; use `ConcurrentHashMap`); `parseOrderFromStatus` (L40-41) aborts the whole parse on a non-string element whereas TS skips it; `humanize`/`compareTo` differ from TS `\b\w`/`localeCompare` on stems with `.`/non-ASCII. All low-impact but genuine.
- **Schema regex:** `readChangeSchema` (`scanner.ts` ~L67, mirrored in `handler.ts`) `^schema:\s*(.+)$` doesn't strip quotes/inline comments and requires column 0; `schema: "spec-driven"  # x` lands verbatim in the badge. Display-only, no functional impact.
- **`action.yml` (~L71):** `npm install -g @fission-ai/openspec` isn't version-pinned, so the static site's schema order depends on whatever's latest at build time; a future CLI output change silently reverts to default order. Degrades safely, but non-reproducible.

### FYI (pre-existing, out of this PR's scope)
`ChangeDetail.tsx` `handleTabChange` uses raw `window.history.replaceState` with react-router's virtual `location.pathname`, which corrupts the real URL under HashRouter (demo) and MemoryRouter (vscode/intellij). This is unchanged from `master` — noting it for a future cleanup, not blocking here.

---
*Verified by running all three suites + repros for findings 1 and 4 against the built `@spek/core`. Findings 2/3 and the Kotlin divergences are from reading the code paths (no live-IDE run).*

kewang:
Heads-up: I cut a couple of patch releases on `master` while this was open (including a `createdDate`/Timeline fix that touched the scanner and the IntelliJ core), so this branch now conflicts with `master` — please rebase/merge before it can land.

Current conflicting files:

- `docs/demo.html` — generated, just rebuild (`npm run build:demo`) to resolve
- `packages/intellij/build.gradle.kts`
- `packages/intellij/src/main/kotlin/com/spek/intellij/core/ChangeReader.kt`
- `packages/intellij/src/main/kotlin/com/spek/intellij/core/Models.kt`
- `packages/intellij/src/main/kotlin/com/spek/intellij/core/OpenSpecScanner.kt`

The Kotlin ones overlap with the same files called out in the review above, so it's worth reconciling both together. Thanks!

nthansen:
Appreciate the feedback! Will start working on that now.

nthansen:
Ready for re-review. Since the last round this branch is now up to date with master (merged in v1.4.0 + the #8 polling fix) and the review findings are addressed, plus a post-merge audit pass on top — the per-commit messages carry the details rather than repeating them here. Happy to address any new feedback.

kewang:
@nthansen — merged, thanks again for the thorough work on this (and the fast turnaround on both review rounds)!

Two small follow-up UX refinements on the schema badge surfaced while testing on a single-schema repo — filed as #9:

1. Show the schema badge in list views (Changes list + Dashboard), not only on Change Detail
2. Hide it when the schema is just the repo default (`spec-driven`), to cut noise

Would be great if you could pick those up since you know this code best — no rush. Thanks!

Issue #2, opened 2026-06-04
Mac安裝完 web版本 npm啟動有問題 [macOS 14.7.3（23H417）]


onlykfc:
~/spek  on master !2  $npm install                                                          ok  system node  at 10:13:51 

up to date, audited 598 packages in 1s

194 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

 ~/spek  on master !2  $npm run dev                                                          ok  system node  at 10:13:59 

----

> spek@1.3.0 dev
> npm run dev -w @spek/web

> @spek/web@0.1.0 dev
> concurrently -n vite,api -c cyan,green "vite" "tsx watch server/index.ts"

[api] 
[api] node:internal/modules/run_main:105
[api]     triggerUncaughtException(
[api]     ^
[api] Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/philwu/spek/node_modules/@spek/core/dist/index.js' imported from /Users/philwu/spek/packages/web/server/routes/openspec.ts
[api]     at finalizeResolution (node:internal/modules/esm/resolve:274:11)
[api]     at moduleResolve (node:internal/modules/esm/resolve:859:10)
[api]     at defaultResolve (node:internal/modules/esm/resolve:983:11)
[api]     at nextResolve (node:internal/modules/esm/hooks:748:28)
[api]     at resolveBase (file:///Users/philwu/spek/node_modules/tsx/dist/esm/index.mjs?1780539253602:2:3744)
[api]     at resolveDirectory (file:///Users/philwu/spek/node_modules/tsx/dist/esm/index.mjs?1780539253602:2:4243)
[api]     at resolveTsPaths (file:///Users/philwu/spek/node_modules/tsx/dist/esm/index.mjs?1780539253602:2:4984)
[api]     at resolve (file:///Users/philwu/spek/node_modules/tsx/dist/esm/index.mjs?1780539253602:2:5361)
[api]     at nextResolve (node:internal/modules/esm/hooks:748:28)
[api]     at Hooks.resolve (node:internal/modules/esm/hooks:240:30) {
[api]   code: 'ERR_MODULE_NOT_FOUND',
[api]   url: 'file:///Users/philwu/spek/node_modules/@spek/core/dist/index.js'
[api] }
[api] 
[api] Node.js v24.4.1
[vite] 上午10:14:14 [vite] (client) Re-optimizing dependencies because lockfile has changed
[vite] 
[...]

Issue #1, opened 2026-03-31
CI github action to use this


mipmip:
This project would be ideal for building a spec site after a release or push to main.

kewang:
Hi @mipmip, thanks for the suggestion! 👋

That's a great idea. We currently have a `build:demo` command that generates a standalone static HTML file (`docs/demo.html`) from an OpenSpec repo. A GitHub Action that automates this on push/release would be a natural next step.

I'm thinking the action could:
1. Take the repo path (or default to the root) as input
2. Run the demo build to generate a static spec site
3. Optionally deploy it to GitHub Pages

Would that match what you have in mind, or are you looking for something different (e.g. validation/linting of OpenSpec content)?

---
_Generated by [Claude Code](https://claude.ai/code)_

mipmip:
Yes this would match it of course. Maybe together with the badges action I published in the action marketplace.

kewang:
Got it — so the main idea is an action to build/publish the spec site, and optionally pair it with badges to show status in the README?

That makes sense. I'll look into it, thanks for the suggestion!

---
_Generated by [Claude Code](https://claude.ai/code)_

kewang:
Hey @mipmip, both features are now live! 🎉

**GitHub Action** — published to the Marketplace:
https://github.com/marketplace/actions/spek-openspec-static-site

Use `kewang/spek@v1` in your workflow to build a static OpenSpec site:

```yaml
- uses: actions/checkout@v4
  with:
    fetch-depth: 0

- uses: kewang/spek@v1
  with:
    title: "My Project - OpenSpec"
```

**Badges** — enable `generate-badges: true` to generate SVG status badges (specs count, open changes, tasks progress):

```yaml
- uses: kewang/spek@v1
  with:
    title: "My Project - OpenSpec"
    generate-badges: true
```

This generates `specs.svg`, `open_changes.svg`, and `tasks.svg` in the output directory. Deploy to GitHub Pages and reference them in your README:

```markdown
![Specs](https://your-user.github.io/your-repo/badges/specs.svg)
![Open Changes](https://your-user.github.io/your-repo/badges/open_changes.svg)
![Tasks](https://your-user.github.io/your-repo/badges/tasks.svg)
```

You can see it in action on spek's own README: https://github.com/kewang/spek

Full docs in the [README](https://github.com/kewang/spek#github-action). Let me know if anything needs adjusting!

PR #8, opened 2026-07-04
fix: watcher polling fallback on network-mounted filesystems (9p/drvfs/NFS/CIFS)

nthansen:
## Why

spek live-reloads while you edit `openspec/`, but its file watchers rely on native OS change events — chokidar→inotify (Web / VS Code) and Java `WatchService`→inotify (IntelliJ). Those events **don't propagate over 9p / drvfs / NFS / CIFS bind mounts** — exactly how devcontainers and WSL mount the host workspace. So files created *after* opening spek (e.g. `brainstorm.md`, `plan.md`, incremental `specs/*/spec.md`) were silently never detected. Works on the host, fails in a container.

## What changed

- **`@spek/core` `watch-polling.ts`** (`shouldUsePolling` / `pollingInterval`): decides via the **watched path's filesystem type** (reads `/proc/mounts`), not "am I in a container" — so a container on `overlayfs` (working inotify) doesn't needlessly poll, while a 9p mount does. Precedence: **explicit override → fstype → remote-env fallback**. Pure decision logic, unit-tested separately from I/O.
- **Web + VS Code (chokidar):** pass `usePolling` / `interval`; VS Code also feeds `vscode.env.remoteName` as a fallback signal.
- **IntelliJ** (`WatchService` has no polling mode): an aligned Kotlin `WatchPolling.kt` plus a polling scan-thread branch in `SpekBrowserPanel`.
- **Override anytime** via `SPEK_WATCH_POLLING=on|off` (Web / VS Code also honor chokidar's `CHOKIDAR_USEPOLLING` / `CHOKIDAR_INTERVAL`).

## Verification

- New `@spek/core` unit tests for the decision logic + IntelliJ Kotlin tests (decision logic + snapshot detection), all passing without a real container. A chokidar polling smoke test confirms a file created after watcher-ready is detected; verified live in an actual 9p devcontainer.
- Type-check + web / webview / vscode builds green.

## Notes

- Split out from a larger PR at the maintainer's request; this is the low-risk half. Contains **no version bump and no CHANGELOG edits** — changelog text below is release-note input only.

### Changelog text (for release notes; not committed)

- Live-reload now falls back to polling on filesystems that don't deliver native change events (9p / drvfs / NFS / CIFS — devcontainers, WSL), so newly created files are detected. Override with `SPEK_WATCH_POLLING=on|off`.


kewang:
## Code review — watcher polling fallback

Thanks for splitting this out — the decision logic is clean, purely functional, and well unit-tested (all 53 `@spek/core` tests green, type-check + web/webview/vscode builds green on my machine). Findings below are ranked by severity; the two top ones I **reproduced at runtime** (repro output inline), the rest are from reading the code.

### 1. `SPEK_WATCH_POLLING=off` is silently ignored on Web & VS Code (chokidar re-reads env) — confirmed, reproduced
`packages/web/server/routes/openspec.ts` (~L38-52) and `packages/vscode/src/watcher.ts` (~L22-39)

`parsePollingOverride` gives `SPEK_WATCH_POLLING` priority and computes `usePolling:false`, but chokidar 5.0.0 re-reads `process.env.CHOKIDAR_USEPOLLING` **after** merging our options and overwrites `opts.usePolling` (`node_modules/chokidar/index.js:284-292`). So if a user has `CHOKIDAR_USEPOLLING=true` exported in their shell rc (the common WSL/Docker workaround) and then sets `SPEK_WATCH_POLLING=off` per the README, Web & VS Code keep polling — only IntelliJ (no chokidar) honors it. This violates the spec scenario "`SPEK_WATCH_POLLING=off` → SHALL NOT poll" and the cross-variant precedence.

```
== Case 1: SPEK_WATCH_POLLING=off, CHOKIDAR_USEPOLLING=true ==
  @spek/core decision usePolling = false   (expected false = no polling)
  chokidar effective usePolling  = true    (overridden by chokidar via env)
  → SPEK_WATCH_POLLING=off defeated, still polling ❌
```

**Fix:** when `SPEK_WATCH_POLLING` is set, reconcile `process.env.CHOKIDAR_USEPOLLING` (delete/normalize it) before calling `watch`, so the core decision is authoritative.

### 2. `CHOKIDAR_INTERVAL` validation is dead on the chokidar surfaces — confirmed, reproduced
`openspec.ts` (~L50) / `watcher.ts` (~L37)

`pollingInterval()` sanitizes `0`/negative/non-numeric to `1000`, but chokidar re-reads raw `CHOKIDAR_INTERVAL` with no `>0` guard (`index.js:294-296`) and overwrites `opts.interval`. `CHOKIDAR_INTERVAL=0` → `fs.watchFile` with interval `0` (busy poll / high CPU) on Web/VS Code, while IntelliJ keeps the validated `1000`.

```
== Case 2: CHOKIDAR_INTERVAL=0 ==
  @spek/core pollingInterval() = 1000   (guard should sanitize 0 → 1000)
  chokidar effective interval  = 0      (overridden by chokidar via env)
  → interval guard defeated, 0 = busy polling ❌
```

### 3. IntelliJ `scanSnapshot` keys only on `lastModified()`, ignoring size — plausible
`packages/intellij/.../core/WatchPolling.kt` (~L147-154)

On coarse-mtime network mounts (exactly the 9p/NFS targets here, often whole-second mtime), an in-place append whose mtime second doesn't advance but whose size grows produces an identical snapshot → no refresh. chokidar's `fs.watchFile` compares the full stat, so Web/VS Code would fire on the same filesystem while IntelliJ wouldn't. Suggest keying on `lastModified + size`.

### 4. IntelliJ teardown race in the polling thread — low
`SpekBrowserPanel.kt` (`setupPollingWatcher` ~L241-263, `dispose` ~L348-354)

`interrupt()` can't abort an in-progress `scanSnapshot`, so a scan that started before `disposed=true` can still call `scheduleRefresh()` → a fresh `Timer()` that runs `executeJavaScript` on the disposed JCEF browser ~500ms later. Mirrors the existing `WatchService` path, so not new behavior, but the new code re-introduces it. A `if (disposed) return` at the top of `scheduleRefresh`/`notifyWebviewFileChanged` closes it.

### Minor / cleanup
- **Duplication:** the chokidar options block (`ignored` predicate + `ignoreInitial/persistent/usePolling/interval/binaryInterval`) is now identical in `watcher.ts` and `openspec.ts` (the vscode file even comments "設定比照 …openspec.ts"). Good moment to extract a shared `buildOpenspecWatchOptions()` into `@spek/core` rather than just the two helper fns.
- **Eager I/O:** `shouldUsePolling` (`watch-polling.ts` ~L161-172) always does `realpathSync` + reads `/proc/mounts` + `existsSync("/.dockerenv")` even when the `SPEK_WATCH_POLLING` override already decides the result (first branch of `decidePolling`). Under aggregation, `watchPaths.some(p => shouldUsePolling(p))` re-reads `/proc/mounts` per path. Check the override first; read `/proc/mounts` once per watcher creation.
- **Release reminder:** the archived `tasks.md` 5.1 is checked ("sync all three CHANGELOGs / 1.4.0 section") but this PR intentionally carries no CHANGELOG edit — just a note to add that `## 1.4.0` section at release time. All three CHANGELOGs are currently identical, so no sync-rule violation.

---
*Verified by running the `@spek/core` suite + a chokidar repro against the installed 5.0.0. IntelliJ findings (3/4) are from tracing the code path, not a live-IDE repro.*

nthansen:
Thanks for the detailed review — the chokidar repro was the key insight. I confirmed the override against the installed chokidar 5.0.0 (`index.js:284-296` re-reads `CHOKIDAR_USEPOLLING`/`CHOKIDAR_INTERVAL` after merging our options), so both #1 and #2 reproduce exactly as you described. Pushed `87466f0` addressing all four findings plus the two actionable minors.

**1. `SPEK_WATCH_POLLING=off` silently ignored (chokidar re-reads env)**
Added `withAuthoritativeChokidarEnv()` in `@spek/core`: it aligns `CHOKIDAR_USEPOLLING`/`CHOKIDAR_INTERVAL` to our computed decision across the *synchronous* `watch()` call, then restores them. Wired into both `openspec.ts` and `watcher.ts`. I went with transient set/restore rather than deleting the env var at startup specifically because the VS Code extension host is a shared process — a permanent mutation could affect other extensions' watchers. This makes the core decision authoritative for the fstype path too, not just the `SPEK_WATCH_POLLING` override.

**2. `CHOKIDAR_INTERVAL` validation dead**
Same helper sets the sanitized interval authoritatively, so `CHOKIDAR_INTERVAL=0` no longer busy-polls. Added a `pollingInterval("0") → 1000` test (mutation testing had flagged it as an uncovered case).

**3. IntelliJ `scanSnapshot` keys only on `lastModified()`**
Now keys on `lastModified + size` (`"mtime:size"`), so an in-place append with an unchanged coarse-second mtime is still detected — matches what `fs.watchFile` does on the chokidar side. Added a test that pins mtime to a fixed second and grows the file.

**4. IntelliJ teardown race**
Guarded `scheduleRefresh` and `notifyWebviewFileChanged` with `disposed`, and set `disposed` inside the same `synchronized(this)` as the debounce cancel in `dispose()`. Putting the check inside the lock (rather than just at the top of `scheduleRefresh`) fully closes the window rather than shrinking it.

**Minor / cleanup**
- *Eager I/O:* `shouldUsePolling` now short-circuits on an explicit override before the `realpathSync` + `/proc/mounts` read.
- *Duplication:* left the two chokidar option blocks as-is rather than extracting a shared helper — they're not quite identical (the VS Code one also watches `addDir`/`unlinkDir`), and I kept this diff focused on the correctness fixes.
- *Release reminder:* noted — no CHANGELOG edit in this PR.

Covered by new `@spek/core` unit tests and mutation testing on `watch-polling.ts` (all mutants on the changed lines killed).

kewang:
Verified the fixes — nice work. The `withAuthoritativeChokidarEnv()` approach (transient env align + restore around the synchronous `watch()`) is a clean way to keep the core decision authoritative without permanently mutating the shared extension-host env.

I ran the two env-override findings against the built `@spek/core` + real chokidar 5.0.0, and both are resolved:

```
== Fixed path (withAuthoritativeChokidarEnv, core decision usePolling=false / interval=1000) ==
  with CHOKIDAR_USEPOLLING=true, CHOKIDAR_INTERVAL=0 in env:
  chokidar effective usePolling = false   ✅ core decision wins, no busy polling
  chokidar effective interval   = 1000    ✅ no longer overwritten by 0
  env after call: CHOKIDAR_USEPOLLING restored to "true"  ✅ process not polluted
```

`@spek/core` suite is green (61/61), and the IntelliJ `scanSnapshot` (`lastModified:size`) + the `dispose()`/`scheduleRefresh()` guard under the shared lock both look correct.

**One thing surfaced by actually building the IntelliJ module, though:** `packages/intellij/build.gradle.kts` on this branch has no `kotlin("test")` dependency, so `WatchPollingTest.kt` (which uses `kotlin.test.*`) doesn't compile — `./gradlew compileTestKotlin` fails with every `kotlin.test` import unresolved:

```
e: WatchPollingTest.kt:5:15 Unresolved reference: test
... (all kotlin.test imports + @Test/assert* usages)
> Task :compileTestKotlin FAILED
```

The `testImplementation(kotlin("test"))` line lives in #6, not here — and rebasing onto `master` won't fix it, since `master` only wires `junit-jupiter` (a different artifact that doesn't provide `kotlin.test.*`). I confirmed both directions: adding `testImplementation(kotlin("test"))` to this branch's `build.gradle.kts` → `compileTestKotlin` passes; swapping in master's junit-jupiter deps instead → still fails. So on its own (this PR is meant to land first), `./gradlew test` / `check` won't compile the new snapshot test.

**Fix:** add `testImplementation(kotlin("test"))` (+ `test { useJUnitPlatform() }` if not already present) to `packages/intellij/build.gradle.kts` in this PR, or port `WatchPollingTest.kt` to the `org.junit.jupiter.api.*` assertions that `master` already provides. (The plugin jar itself still builds; it's `compileTestKotlin`/`test` that breaks.)

nthansen:
Good catch, and apologies for the miss — I'd earlier written the `compileTestKotlin` failure off as a local toolchain issue when it was actually our own gap: this branch's `build.gradle.kts` wired no test framework, so the `kotlin.test`-based `WatchPollingTest` couldn't compile. Fixed in `8ffa6a3` by adding `testImplementation(kotlin("test"))` + `test { useJUnitPlatform() }`, and verified with `./gradlew test` that WatchPollingTest's 22 tests (incl. the new mtime/size snapshot case) now compile and pass.

Issue #9, opened 2026-07-07
Schema badge: show it in list views, and hide it when it's the repo default


kewang:
Follow-up to #6 (custom OpenSpec schema support) — two small UX refinements on the schema badge. @nthansen, would you be up for taking these since you built the feature?

## 1. Show the schema badge in list views, not only on Change Detail

Today the badge renders only on the Change Detail page (`ChangeDetail.tsx`). The Changes list (`ChangeList.tsx`) and the Dashboard don't show it, so you can't tell a change's schema at a glance — you have to open each change one by one. But spotting a differently-shaped change in a mixed-schema repo is exactly a *list*-level need, so that's where the badge earns its keep.

- Add the badge to each row in `ChangeList.tsx` — there's already a `WorktreeBadge` in that row, so it can sit alongside it.
- Surface it on the Dashboard too, wherever changes are represented.
- `ChangeInfo` already carries `schema`, so the list data is already there (no API change needed).

## 2. Hide the badge when it's just the repo default (reduce noise)

In a single-schema repo — the common case, where everything is `spec-driven` — every change shows an identical `spec-driven` badge, which is pure noise. Hide the badge when the change's schema is `spec-driven`.

- Design note: consider keying this off the repo default in `openspec/config.yaml` (hide when `change.schema === repoDefaultSchema`) rather than hardcoding the string `spec-driven`. That way a change authored under an unusual schema still stands out even in a spec-driven repo, and a repo whose default is some *other* schema also gets a clean list. Your call on which reads cleaner — hardcoding `spec-driven` is simpler; the config-default comparison is more general.

## Net effect

In a plain spec-driven repo the list stays clean (no badges at all), and the moment a change uses a different schema its badge pops right there in the list — which is precisely when the badge is useful.

nthansen:
Yup! Happy to take this on later.

kewang:
Done — implemented by @nthansen in #10 (merged) and shipped in [v1.5.0](https://github.com/kewang/spek/releases/tag/v1.5.0).

Both requested items landed, plus a `Default schema: <name>` label on the Changes page that explains the (now badge-less) baseline. The hide rule is keyed off the repo default in `openspec/config.yaml` rather than a hardcoded `spec-driven`, so a repo whose default is a custom schema gets a clean list too — verified by flipping this repo's config and watching the badges invert.

A follow-up optimization surfaced during review and is tracked separately in #11.

PR #10, opened 2026-07-08
Show schema badge in list views and hide it when it matches the repo default (#9)

nthansen:
Addresses #9 — the two schema-badge refinements, plus one small addition and a couple of judgment calls flagged below for your sign-off.

## The two requested items

1. **Badge in list views.** Extracted the inline badge from Change Detail into a reusable `SchemaBadge`, rendered on the Changes list rows and the Dashboard rows (not only Change Detail).
2. **Hide when it's the repo default.** The badge is hidden whenever a change's schema equals the repo default and only shows when a change diverges — applied uniformly across Change Detail, the Changes list, and the Dashboard.

The repo default is threaded to the frontend as a new `defaultSchema` field on `ScanResult` / `ChangesData` / `ChangeDetail`, read directly from `openspec/config.yaml` (no CLI, off the scan hot path), and mirrored in the IntelliJ Kotlin core.

## Design choices to sign off on

Easy to adjust or drop any of these:

1. **Config-default comparison over hardcoding `spec-driven`.** You left this to my call — I keyed the hide rule off `openspec/config.yaml`'s `schema:` (hide when `change.schema === repoDefaultSchema`), so a repo whose default is a custom schema also gets a clean list. When no default is determinable, `defaultSchema` is `null` and the badge simply shows.
2. **A plain-text `Default schema: <name>` label on the Changes page** — the one thing beyond the issue's two items. Because the hide rule suppresses the badge on every default-schema change, a badge-less list is otherwise ambiguous about what everything shares; this states the baseline right where the divergent badges appear. It's plain text (not a pill) so the pill stays an unambiguous "this one is non-default" signal, and it reads `ChangesData.defaultSchema` (already fetched — no extra request). **Happy to drop it if you'd rather keep the PR to exactly the two issue items.**
3. **Row badge placement.** The issue suggested the badge sit alongside `WorktreeBadge` by the title; I put it on the right of the row with the date/lifecycle metadata to keep the change title clean. `WorktreeBadge` stays exactly where you had it. Trivial to move back.
4. **Uniform hide includes Change Detail.** The detail header previously always showed the badge; it now hides there too when the schema equals the default, for consistency with the lists.

## Known limitation

In an aggregated cross-worktree view, `ChangesData.defaultSchema` is the main worktree's config while each change's schema resolves against its own worktree's config, so with *divergent per-worktree* `config.yaml` schemas a change could show a badge in the list but not on its detail page. Rare (config.yaml is committed and normally shared); documented in the change's `design.md` rather than fixed, to avoid per-change default plumbing for an edge case.

## Testing

- `@spek/core` + `@spek/web` unit tests (badge predicate; `defaultSchema` on single/aggregated scans and `ChangeDetail`; config-default comparison).
- IntelliJ Kotlin core tests for `defaultSchema` parity.
- `npm run type-check` and `npm run build` green; IntelliJ Gradle tests green.

The OpenSpec change (proposal / design / spec delta / tasks) is included under `openspec/changes/schema-badge-refinements/`.


nthansen:
Marking as draft while I archive the OpenSpec change — this folds the spec delta into the main `custom-schema-artifacts` spec and moves the change record to `openspec/changes/archive/`, per the repo's OpenSpec workflow. I'll push that and mark it ready again shortly.

kewang:
Reviewed and merging — nice work. **All four design calls are signed off as-is**, no changes needed:

1. **Config-default comparison over hardcoding `spec-driven`** — right call, and I verified it behaves. I flipped this repo's `openspec/config.yaml` to `schema: superpowers-bridge` against a scratch pair of changes: the `spec-driven` change's badge appeared and the `superpowers-bridge` one's disappeared, and the `Default schema:` label followed. So it genuinely compares against the config default rather than a hardcoded string, and a repo whose default *is* a custom schema gets a clean list.
2. **Keep the `Default schema: <name>` label.** Your reasoning holds: once the hide rule suppresses every default badge, a badge-less list is otherwise mute about what everything shares. Plain text rather than a pill is the right distinction — it keeps "pill = this one is non-default" as an unambiguous single signal.
3. **Badge on the right of the row** — better than what the issue suggested. Titles stay clean, the schema reads as row metadata alongside the date, and `WorktreeBadge` stays put.
4. **Uniform hide including Change Detail** — correct, that's the consistency I wanted.

## Verification

Ran locally on the branch: `@spek/core` 115, `@spek/web` 36, IntelliJ Kotlin 58 — all green; `type-check` and `build` clean. Traced `defaultSchema` through all four assemblers (web route, VS Code handler, `build-demo.ts`, Kotlin handler) and confirmed the Kotlin `readRepoSchema` regex + `cleanScalar` are equivalent to the TS side. Also checked the `SchemaBadge` predicate across the null/undefined/empty-string matrix — the `defaultSchema == null` fail-open is benign on every path I could reach (missing `config.yaml`, indented `schema:` key, StaticAdapter), and there's no load-time badge flash since `defaultSchema` and the rows come from the same query.

Extracting the pill into one `SchemaBadge` with the hide rule in exactly one place is the right shape — thanks for that.

## Two notes, neither blocking

- **The aggregated baseline asymmetry you documented in `design.md`** — I filed it as #11 together with a related efficiency issue, because one fix covers both. `readChangeSchema`'s fallback re-reads `config.yaml` once per change (N+M+1 reads per scan), so threading the already-computed `defaultSchema` into `scanChangeDir` and stamping it onto `ChangeInfo` collapses that to a single read *and* makes each change compare against its own worktree's default — which removes the false-positive badge and the list↔detail mismatch. The plumbing you (reasonably) judged not worth it for a rare case pays for itself once the hot-path I/O is counted. Happy for that to be a separate PR whenever.
- **`docs/demo.html` wasn't rebuilt**, and `pages.yml` deploys `docs/` verbatim with no build step — so the live demo will keep showing the pre-#10 UI until `npm run build:demo` runs. That matches your "bundles rebuilt at packaging/release" convention and there's no drift risk (demo.html bundles app + data together), just a note to self for the next release.

PR #13, opened 2026-07-11
Thread defaultSchema into scanChangeDir: collapse config.yaml reads and fix aggregated baseline (#11)

### LINKED ISSUE #11: Thread defaultSchema into scanChangeDir: collapse N+M config.yaml reads and fix the aggregated baseline asymmetry
Follow-up to #10. One change fixes two things at once — an efficiency issue on the changes hot path, and the aggregated cross-worktree asymmetry that #10 documented as a known limitation.

## The two problems

**A. `openspec/config.yaml` is read N+M+1 times per scan.**

`readChangeSchema` (`scanner.ts:83-89`) falls back to `readRepoSchema(repoDir)` whenever a change doesn't declare its own `schema:` in `.openspec.yaml` — which is the common case. So `scanOpenSpec` does one `existsSync` + `readFileSync` + regex of the *same* `config.yaml` for each of the N active + M archived changes (`scanner.ts:128`), and #10 adds one more for `ScanResult.defaultSchema` (`scanner.ts:191`). Multiply by worktree count under aggregation. The file is small so the cost is low, but it's repeated I/O on the `/changes` and `/overview` path that's trivially avoidable.

**B. Aggregated view compares each change against the *wrong* default.**

`scanOpenSpecAggregated` ships a single `defaultSchema` taken from the main worktree (`scanner.ts:465`), but each change's `schema` was resolved against *its own* worktree's `config.yaml`. With divergent per-worktree `config.yaml` schemas:

- A change in worktree B that declares **no** schema inherits B's default → `c.schema = B_default`. The list compares it against the *main* default → a badge appears, falsely implying the change opted into a non-default schema.
- Opening that change's detail resolves `readChange(B_path, …)` → `defaultSchema = B_default` → badge hidden. So the same change shows a badge in the list but not on its detail page.

#10 documented this in `design.md` (Risks) and chose not to fix it, reasonably, to avoid "per-change default plumbing for a rare case."

## The fix (one change, both problems)

Thread the already-computed default into the per-change scan, and carry it on `ChangeInfo`:

1. `scanOpenSpec` computes `const defaultSchema = readRepoSchema(repoDir)` **once** and passes it to `scanChangeDir(repoDir, changePath, slug, status, defaultSchema)`.
2. `readChangeSchema` takes that value instead of re-reading — collapsing N+M+1 disk reads to **1**.
3. `scanChangeDir` stamps `defaultSchema` onto `ChangeInfo`. In aggregated mode each worktree's scan stamps *its own* default, so every change carries the right baseline.
4. The list/dashboard render `<SchemaBadge schema={c.schema} defaultSchema={c.defaultSchema} />` — per-change comparison, so the false-positive badge and the list↔detail mismatch both disappear.
5. `ChangesData.defaultSchema` stays as-is and keeps feeding the `Default schema: <name>` page label — that's the main-repo baseline statement, which is still exactly the right thing for the header.

The plumbing the design doc wanted to avoid turns out to pay for itself: it removes N+M redundant reads on the hot path *and* makes the aggregated comparison correct. Mirror the same in the Kotlin core (`OpenSpecScanner.kt` has the identical `readChangeSchema` fallback pattern; IntelliJ has no aggregation path, so only the efficiency half applies there).

Low priority — neither problem bites a normal single-worktree, single-schema repo. Filing so it isn't lost.

nthansen:
Closes #11. Follow-up to #10.

One change fixes the two problems #11 describes, both rooted in re-resolving the repo default schema per change.

## A. Efficiency — `openspec/config.yaml` read once, not N+M+1

`readChangeSchema` fell back to `readRepoSchema(repoDir)` for every change without its own `schema:` (the common case). `scanOpenSpec` now computes `readRepoSchema(repoDir)` **once** and threads it into `scanChangeDir`; `readChangeSchema(changePath, defaultSchema)` takes the value instead of re-reading. `readChange` no longer reads `config.yaml` twice either. N+M+1 reads → 1.

## B. Correctness — aggregated view compares against the right default

Each `ChangeInfo` now carries a `defaultSchema` field, stamped from the worktree it was scanned in. Because `scanOpenSpecAggregated` already scans each worktree with its own `repoDir`, every change now carries **its own worktree's** baseline — so the list/dashboard badge compares each change against its own default. This removes the false-positive badge and the list↔detail mismatch for divergent per-worktree `config.yaml` schemas that #10's `design.md` documented as a known limitation.

`ChangesData.defaultSchema` is unchanged and still feeds the `Default schema: <name>` page header (the main-repo baseline, which is correct there).

## Changes

- **core** (`scanner.ts`, `types.ts`): thread `defaultSchema`, add it to `ChangeInfo`, per-change stamping.
- **web UI** (`ChangeList`, `Dashboard`): `<SchemaBadge defaultSchema={c.defaultSchema} />` (per-change); `ChangeDetail` and the Changes header untouched. Adapters pass the field through unchanged.
- **Kotlin core** (`OpenSpecScanner.kt`, `ChangeReader.kt`): mirror the read-once fallback. IntelliJ has no aggregation path, so only the efficiency half applies there.

No behavior change for single-worktree, single-schema repos.

## Tests

- Core (`scanner.test.ts`, `aggregate.test.ts`): per-change `defaultSchema` stamping, **read-once** (spy asserts exactly one `config.yaml` read), and a **divergent-worktree** aggregation case (a no-schema change in worktree B carries B's default, not main's).
- Kotlin (`OpenSpecScannerTest.kt`): asserts each `ChangeInfo.defaultSchema` is stamped.

Verified locally: `type-check` clean, core tests green, `build` + `build:demo` succeed, Kotlin tests pass. Mutation testing on the changed scanner logic: 12/12 mutants killed.

Developed via the OpenSpec workflow; the archived change (proposal/design/specs/tasks) and the synced `custom-schema-artifacts` spec are included.


kewang:
Reviewed. The core fix is right and I verified both of its claims against a real fixture — a git repo with a second worktree whose `config.yaml` declares a different default:

```
A. config.yaml reads (6 changes):  1  ✅  (was N+1 = 7)

B. main=spec-driven, B=superpowers, neither change declares its own schema:
   b-change  schema=superpowers  defaultSchema=superpowers  → badge hidden ✅ (false positive gone)
   b-change  list=hidden  detail=hidden                     → consistent   ✅
   reverse: b-change declares `schema: weird-schema`        → badge shown  ✅ (no false negative)
```

Read-once, per-worktree baseline, list↔detail agreement, and the reverse case all behave. Nice, tight change — no missed call sites, the aggregate spread preserves the stamp, and the `mock.method(fs, "readFileSync")` spy is a genuine regression guard rather than a no-op.

**I've pushed three follow-up commits to this branch** rather than bouncing them back to you, since they're mechanical. Details below — please shout if you'd rather I'd left any of them to you.

## 1. The required field broke `@spekjs/ui` (this was the blocker)

`ChangeInfo` is public API of the **published** `@spekjs/core`, and `@spekjs/ui`'s `ChangeTimeline` takes `changes: ChangeInfo[]` — components are pure presentation, data comes in via props, so hosts *construct* `ChangeInfo`. Making `defaultSchema` required is therefore source-breaking for constructors.

It already broke inside this repo:

```
packages/ui/src/timeline/__tests__/grouping.test.ts(7,3):
  error TS2741: Property 'defaultSchema' is missing ... but required in type 'ChangeInfo'.
```

Zero such errors on `master` — so this PR introduced it. It stayed invisible because `packages/ui/tsconfig.json` has `"exclude": ["src/**/__tests__/**"]` and the suite runs through tsx with no typecheck: **22 ui tests green on a type-invalid file.**

Fixed the fixture, and bumped `@spekjs/core` to **1.1.0** with the breaking type change documented prominently in the package CHANGELOG (migration note + the reason it ships as a minor). Semver would normally say `2.0.0` here; kewang's call is to ship it as a minor because the only known consumer is his own downstream app, which he's updating in step — but it's called out loudly rather than smuggled through. `design.md`'s "no breaking changes" line is what tipped me off to check.

## 2. Kotlin lost the compile-time guard (`Models.kt`)

`val defaultSchema: String? = null` on `ChangeInfo` / `ChangeDetail` / `ChangesData`. The whole point of making the TS field required is that a producer which forgets to stamp it fails to compile — but a Kotlin default reinstates exactly the silent failure: it compiles, `encodeDefaults = true` emits `"defaultSchema": null`, and the shared frontend's `SchemaBadge("spec-driven", null)` → `false` → **a badge on every change in a single-schema repo**, i.e. the noise this whole feature exists to suppress.

Dropped the defaults. All three construction sites already passed the value, so it compiles clean — Kotlin suite is 73/73.

## Verified after the fixes

`@spekjs/core` 118 · `@spekjs/web` 14 · `@spekjs/ui` 22 · IntelliJ Kotlin 73 — all green; `type-check` and `build` clean; the `TS2741` is gone.

## Two things I deliberately did *not* change

- **The header/badge basis now diverges under aggregation.** Rows compare against their own worktree default, but the `Default schema: <name>` header still shows the *main* worktree's. In a divergent-config setup a change whose schema genuinely differs from the displayed repo default now shows **no badge at all** — the old false positive traded for a false negative, and cross-worktree schema divergence becomes invisible. It's rare (needs divergent committed `config.yaml`s) and kewang is fine leaving it as a known state, but flagging it since your `design.md` blessed the per-change basis without revisiting the header.
- **The new aggregate test leans on enumeration order.** Worktree B branches from a commit that already contains `main-change`, so the union yields *two* rows with that slug and `find()` picks the main one only by V8's stable sort plus `listWorktrees` returning main first. It also never asserts `activeChanges.length`, so it green-lights the duplicate-row/no-badge case above instead of pinning it. Consider `filter()` + select by `source.key`, plus a length assertion.

Separately: there's no test/typecheck CI in `.github/workflows/` at all, and `npm run type-check` only covers `packages/web` — which is precisely why a type-invalid file could sit in `packages/ui` unnoticed. Worth its own issue.

Merging once you've had a look at the pushed commits.

PR #14, opened 2026-07-11
fix(toc): land clicked heading below the sticky header, not one section too far

nthansen:
## Why

Clicking a TOC entry (or opening a deep-link `#hash`) on the **Change detail** and **Spec detail** pages scrolled the target heading *behind* the sticky header, so the heading was hidden and the **next** section showed at the top — it looked like the click jumped one section too far. The scroll handlers hardcoded an 80px offset, but the Change detail sticky header (title + tab row) is taller than that.

## What changed

- New `scrollOffset` util (`packages/web/src/utils/scrollOffset.ts`) measures the real sticky-header bottom at scroll time: content sticky header (`[data-spek-scroll-offset]`) → fixed app header (`header.fixed`) → constant fallback. Its `anchorScrollTop` computation is pure and unit-tested.
- `ChangeDetail` + `SpecDetail` hash-anchor effects and the `useScrollspy` active-heading tracking all consume it, so both the scroll target and the highlighted TOC entry track the actual header height.
- `TabView` marks the sticky tab-row container with `data-spek-scroll-offset`.

No core/API/extension changes; no behavior change beyond the scroll landing position.

## Testing

- `packages/web/src/utils/scrollOffset.test.ts` — unit tests `anchorScrollTop` and the `scrollOffset` fallback chain (4 tests).
- `npm run type-check` clean; `npm run build` succeeds.
- Verified in the running app on a Change detail with a tall sticky header: a clicked TOC heading now lands just below the sticky bar instead of behind it.

Developed via the OpenSpec workflow; the archived change and the synced `change-browsing` / `spec-browsing` hash-anchor-navigation requirements are included.


kewang:
Thanks Norman — the root fix (measuring the header instead of assuming 80px) is the right call, and the tall `ChangeDetail` sticky header was exactly the trap. I reviewed it and found one thing left over, which I have pushed a follow-up commit for (430e085) rather than sending you round another loop, since it hinges on layout geometry that is easy to miss.

## What was still off

`scrollOffset()` measured the sticky header via `rect.bottom` at scroll time. While the page sits at the very top the header has not pinned yet — `main` has `pt-18` (72px) while the container is `sticky top-14` (56px) — so the measurement comes out 16px lower than where the header actually ends up. Anchor scrolling and the scrollspy threshold then disagree by that 16px. And because an anchor scroll parks the target heading exactly *on* the threshold (both now derive from the same `scrollOffset()`), sub-pixel rounding can flip the comparison even mid-scroll. The old code happened to have a 16px cushion between the two constants (80 vs 96) which absorbed this; unifying them removed it.

Either way the heading lands correctly, but the TOC keeps highlighting the entry *before* the one the reader clicked.

Measured in the running app on a change detail with a 169px sticky header (17 headings), before vs after:

| start | clicked | highlighted (before) | highlighted (after) |
|---|---|---|---|
| page top | D6 | **D5** (heading top 253.5, threshold 237.3) | D6 |
| mid-scroll | Risks / Trade-offs | **D11** (heading top 237.5, threshold 237.3) | Risks / Trade-offs |

The two rows are the two causes: the 16px pin drift, and the sub-pixel boundary.

## The follow-up commit

- `pinnedBottom(rect, cssTop)` — derive the obstruction bottom as it sits *once pinned* (CSS `top` + height, capped by `rect.bottom` so a header scrolling away with its parent still measures right). The offset no longer drifts with the reader's scroll position, which also makes the SpecToc double-scroll (click scrolls, hash effect scrolls again) idempotent. Pure, unit-tested.
- `activeHeadingId(headings, threshold)` — the scrollspy pick, extracted and unit-tested, with a few px of tolerance for rounding.
- App header selected via `data-spek-app-header` rather than `header.fixed`: a Tailwind utility class in a selector fails silently if the class ever changes, and you already had the right pattern with `data-spek-scroll-offset`.
- Folded into the change's existing artifacts (tasks + a `The clicked entry is the highlighted one` scenario on both scrollspy requirements) instead of opening a second change, and restored some blank-line churn the spec sync introduced.

`npm run type-check` clean, `npm test -w @spekjs/web` green (24), `npm run build` succeeds.

Separately, and **not** something this PR introduced: headings near the end of a document can never reach the threshold line because `scrollTo` is clamped by the document height, so the last TOC entry or two never light up. Pre-existing, worth its own change.

Issue #4, opened 2026-06-18
Plugin incompatible with IntelliJ Platform 2026.1 (build 261.x) — "requires IDE build 253.* or earlier"


tuchorc:
# Describe the bug

The "spek - OpenSpec Viewer" plugin cannot be installed/enabled on current 2026.1.x JetBrains IDEs. The IDE reports:

> Incompatible: requires IDE build 253.* or earlier

This suggests the plugin's `until-build` constraint in `plugin.xml` hasn't been updated past the 253.x branch, so it's being blocked on newer platform versions even though there's likely no actual incompatibility.
Environment

IntelliJ IDEA 2026.1.3, Build #IU-261.25134.95 (built June 3, 2026)
WebStorm 2026.1.3, Build #WS-261.25134.101 (built June 4, 2026)
Plugin: spek - OpenSpec Viewer, latest version from JetBrains Marketplace (Plugin ID `tw.kewang.spek`)

# Steps to reproduce

Open IntelliJ IDEA 2026.1.3 or WebStorm 2026.1.3
Go to Settings → Plugins → Marketplace
Search for "spek - OpenSpec Viewer" and attempt to install (or try installing the latest version manually)
IDE blocks installation/enabling with the "requires IDE build 253.* or earlier" message

# Expected behavior

The plugin should install and run normally on current IDE builds (261.x), consistent with the VS Code extension, which appears to be actively maintained and updated (latest npm/marketplace release dated June 2026).

#Additional context

- Repo: https://github.com/kewang/spek
- The VS Code and Open VSX listings for spek seem up to date, so this looks like the JetBrains plugin's `until-build` value in `plugin.xml` / `build.gradle.kts` simply needs bumping to track newer IntelliJ Platform releases — possibly an oversight rather than a real compatibility issue.

-------
Generated by [Claude Code](https://claude.ai/code)

kewang:
Fixed in v1.3.3.

The plugin's `until-build` upper bound (`253.*`) has been removed:
- `packages/intellij/gradle.properties` no longer sets `pluginUntilBuild`.
- `build.gradle.kts` sets `ideaVersion.untilBuild = provider { null }`, so the patched `plugin.xml` now emits `<idea-version since-build="233" />` with **no** upper bound.

[...]

Issue #3, opened 2026-06-04
`npm run dev` fails on a fresh clone: @spek/core is never built, Vite can't resolve @spek/core/headings


onlykfc:
## Summary

Following the **Web Version** Quick Start on a fresh clone fails immediately. The Vite dev server cannot resolve the `@spek/core/headings` subpath, because `@spek/core` is never built before `npm run dev` starts.

## Steps to reproduce

Exactly the documented Quick Start (`README.md` → Quick Start → Web Version):

```bash
git clone https://github.com/kewang/spek.git
cd spek
npm install
npm run dev
```

Open http://localhost:5173 → Vite shows an error overlay:

```
[plugin:vite:import-analysis] Failed to resolve import "@spek/core/headings"
from "src/pages/ChangeDetail.tsx". Does the file exist?

20 |  import { extractHeadings } from "@spek/core/headings";
```

Reproduces 100% of the time on a fresh clone (and after any `dist/` clean).

## Root cause

- `packages/web/src/pages/ChangeDetail.tsx` imports from the `@spek/core/headings` subpath.
- `packages/core/package.json` maps `exports["./headings"]` → `./dist/headings.js` — i.e. the **compiled** output:

  ```json
  "exports": {
    "./headings": { "import": "./dist/headings.js", "types": "./dist/headings.d.ts" }
  }
  ```

- `npm install` does **not** build internal workspace packages, and the root `dev` script is just `npm run dev -w @spek/web`, whose script only runs Vite + `tsx watch` for the web package — it **never builds `@spek/core`**.
- So on a fresh clone `packages/core/dist/` does not exist, and Vite cannot resolve the subpath.

The documented Quick Start (`npm install && npm run dev`) therefore cannot work as written.

## Workaround

Build core once before starting the dev server:

```bash
npm run build:core   # = npm run build -w @spek/core (runs tsc, produces dist/)
npm run dev
```

## Suggested fix

Make `npm run dev` build core first (or watch it in parallel) so the documented Quick Start works with no extra manual step. For example, the root `dev` script:

```jsonc
// option A: build once before starting web dev
"dev": "npm run build -w @spek/core && npm run dev -w @spek/web"
```

or run `tsc --watch` for core alongside Vite via `concurrently`, so core changes also rebuild during development. Either keeps `npm install && npm run dev` accurate for new contributors.

## Environment

- Node v24.4.1, npm 11.13.0
- Vite 6.x
- macOS 14.7.3

onlykfc:
[...]

PR #17, opened 2026-07-12
fix: dedupe active changes across worktrees

david-lutz:
## Summary

A worktree forked from `main` inherits **every** open change directory, not just the one it's editing. Cross-worktree aggregation unioned active changes with no dedup, so a change present in N worktrees showed N times in both the Changes list and the Graph view, and duplicate graph edges inflated spec fan-in (`historyCount`).

This dedupes by slug, electing the copy that is **actually being worked on**.

## Election: git divergence, not mtime

The election signal is **git divergence** — which copy has advanced its change past the point it was forked from — replacing the filesystem-mtime signal from the first revision of this PR.

For a given slug, the **candidates** are the copies that have advanced past their merge-base:

- A non-main worktree diverges when its `HEAD` advances `openspec/changes/<slug>/` beyond the merge-base with `main` — the **three-dot** `git diff --name-only <mainHead>...<wtHead>`, which counts only the worktree's *own* post-fork commits — **or** it has uncommitted modifications there (`git status --porcelain`).
- **`main` competes on the same terms** — no special case. When the contest is live (some worktree diverges on the slug) and `main` has itself advanced past that worktree's merge-base (the reverse three-dot `git diff <wtHead>...<mainHead>`) or has uncommitted edits there, `main` joins the candidate pool.
- When **no** copy diverges, `main` keeps the slug.
- Among the candidates, the copy whose change directory was **modified most recently** (mtime) wins the tiebreak.
- A per-command git failure treats that copy as not diverging, mirroring `git-cache.ts`'s resolve-empty-on-error.

The winning copy is the one whose *contents* the list and Dashboard render — `taskStats` included — so the election decides what you see, not just a `source` label. The same helper (`pickActiveWinners`) drives both the Changes list and the Graph over the **same slug set**, so the two can't disagree about who owns a change.

### Behaviour by scenario

For a slug shared between `main` and a worktree, forked at `F`:

| Situation | Winner |
|---|---|
| Worktree advanced past `F`; `main` didn't | worktree |
| `main` advanced past `F`; worktree only inherited it | `main` |
| **Both** advanced past `F` | most recently modified copy (either can win) |
| Two worktrees advanced; `main` didn't | most recently modified worktree |
| Nobody advanced (all inherited) | `main` |

### Why mtime was wrong

`git worktree add` writes every file in the new tree fresh, so mtime measures **checkout time, not edit time**. The most-recently-created worktree won every slug it merely inherited, and an idle fork permanently shadowed `main`. Because the winning copy is the one rendered, task progress silently rolled back to the fork-point snapshot (a 3/4 change reported as 0/4). Both original repro scenarios now pass:

| Scenario | Result |
|---|---|
| `wa` edits `change-a` to 3/4, `wb` created afterwards inheriting it | `wa`, **3/4** |
| `main` edits `change-x` to 4/4 (uncommitted), `wb` created afterwards inheriting it | `main`, **4/4** |

The **three-dot** diff also closes a subtler variant: a **two-dot** `git diff <mainHead> <wtHead>` fires whenever the endpoints differ on a slug — including when *`main`* advances a slug an idle worktree merely inherited — which would elect the idle fork's stale copy. Three-dot (merge-base–relative) counts only the worktree's own advances.

## Addressed in review

- **Signal**: filesystem mtime → git divergence; the "non-main unconditionally beats main" carve-out is gone (`main` competes on equal terms, so a finished copy on `main` can't be shadowed by a stale-but-edited fork, or vice-versa).
- **Comparison** is against `main`'s `HEAD` (not its working tree), and via three-dot (not two-dot), so an inherited-but-untouched copy is never a candidate whichever side advanced.
- **Tests** use real `git worktree add` in natural creation order (no `fs.utimesSync` except where a test specifically exercises the mtime tiebreak among genuinely-diverging copies) and assert both the winning `source` **and** `taskStats`, so a silent content rollback fails the suite.
- **Graph path** derives its election slug set from the same source as the list (each worktree's `changes/` dir), so list and Graph elect the same winner even for a change whose delta spec exists in only one worktree; it no longer runs a `scanOpenSpec` per worktree just to obtain slugs.
- **Docs**: `pickActiveWinners` / `scanOpenSpecAggregated` / `buildGraphDataAggregated` docstrings, `CLAUDE.md`, and CHANGELOGs (`@spekjs/core` + root) describe the divergence election.

## Notes

- No API/frontend contract changes — same `ChangeInfo.source` and `change:<worktreeKey>:<slug>` node id; only which entries survive dedup changes.
- The OpenSpec change and its `proposal` / `design` / `specs` / `tasks` present divergence as the design, with mtime demoted to a rejected primary-signal alternative in `design.md`.


kewang (changes requested):
Thanks for this — the diagnosis is right and the problem is worth fixing. A worktree forked from `main` really does inherit every open change directory, the union really does produce a wall of duplicates, and the inflated `historyCount` from duplicate graph edges is a real bug. The OpenSpec artifacts are thorough, and I appreciate that `design.md` honestly records the mid-implementation correction about `ChangeInfo.timestamp` being the *earliest* commit. Sharing one election helper between the list and graph paths so they can't disagree is the right structural call.

But I can't merge this as it stands: **the election signal is wrong, and it fails precisely in the workflow this PR sets out to fix.**

---

## Blocking: filesystem mtime measures checkout time, not edit time

`git worktree add` writes every file in the new working tree fresh. So a change directory that a worktree merely *inherited and never touched* carries an mtime equal to that worktree's creation time. The consequence: **the most recently created worktree wins every slug it inherited**, despite never having touched any of them.

Combined with the unconditional "non-main always beats main" rule, an idle worktree also permanently shadows work that is happening on `main` itself.

I reproduced both against this branch, with real git repos and real `git worktree add` (no `utimesSync`):

| Scenario | Reality | This PR's result | Expected |
|---|---|---|---|
| `wa` is working on `change-a` (3/4 tasks done); `wb` is created afterwards for a *different* change | `change-a` is being edited in `wa` | source = **`wb`**, taskStats **0/4** | `wa`, 3/4 |
| `main` edits `change-x` to 4/4 (uncommitted); a worktree `wb` is created afterwards | `change-x` is being edited on `main` | source = **`wb`** (`isMain: false`), taskStats **0/4** | `main`, 4/4 |

This is not just a mislabelled `source`. The winning copy is the copy whose *contents* the list renders, so **Changes and the Dashboard silently roll task progress back to the fork-point snapshot**. A contributor sees a change they are halfway through reported as 0/4. For a change that is finished and awaiting review — nobody is touching its files any more — the wrong copy wins *permanently*.

Against the status quo this is a net regression in data correctness: today's output is noisy but at least contains a correct row; after this change there is exactly one row and it can be the stale one.

Repro script (run from a checkout of this branch, `npx tsx repro.mts`):

```ts
import fs from "node:fs";
import os from "node:os";
import path from "node:path";
import { execFileSync } from "node:child_process";
import { scanOpenSpecAggregated } from "./packages/core/src/scanner.js";

const env = { ...process.env, GIT_AUTHOR_NAME: "t", GIT_AUTHOR_EMAIL: "t@t", GIT_COMMITTER_NAME: "t", GIT_COMMITTER_EMAIL: "t@t" };
const git = (cwd: string, ...a: string[]) => execFileSync("git", a, { cwd, stdio: "pipe", env });
const w = (p: string, c: string) => { fs.mkdirSync(path.dirname(p), { recursive: true }); fs.writeFileSync(p, c); };
const commit = (r: string, m: string) => { git(r, "add", "-A"); git(r, "commit", "-q", "-m", m); };
const tasks = (done: number, total: number) =>
  "## 1. T\n" + Array.from({ length: total }, (_, i) => `- [${i < done ? "x" : " "}] 1.${i + 1} step\n`).join("");
const initRepo = (prefix: string) => {
  const r = fs.mkdtempSync(path.join(os.tmpdir(), prefix));
  git(r, "init", "-q", "-b", "main");
  w(path.join(r, "openspec", "config.yaml"), "schema: spec-driven\n");
  return r;
};

// Scenario 1: wa is editing change-a; wb is created later for change-b
{
  const repo = initRepo("s1-");
  for (const s of ["change-a", "change-b"]) {
    w(path.join(repo, "openspec", "changes", s, "proposal.md"), "## Why\nx\n");
    w(path.join(repo, "openspec", "changes", s, "tasks.md"), tasks(0, 4));
  }
  commit(repo, "init: two open changes on main");

  const wtA = repo + "-a";
  git(repo, "worktree", "add", "-q", "-b", "wa", wtA);
  w(path.join(wtA, "openspec", "changes", "change-a", "tasks.md"), tasks(3, 4)); // wa really works on change-a
  commit(wtA, "a: progress on change-a");

  const wtB = repo + "-b";
  git(repo, "worktree", "add", "-q", "-b", "wb", wtB); // wb inherits change-a with a fresh mtime

  const a = (await scanOpenSpecAggregated(repo)).activeChanges.find((c) => c.slug === "change-a")!;
  console.log("[1] winner:", a.source?.branch, JSON.stringify(a.taskStats), "— expected wa / 3-of-4");
}

// Scenario 2: main is editing change-x (uncommitted); a worktree is created afterwards
{
  const repo = initRepo("s2-");
  w(path.join(repo, "openspec", "changes", "change-x", "proposal.md"), "## Why\nx\n");
  w(path.join(repo, "openspec", "changes", "change-x", "tasks.md"), tasks(0, 4));
  commit(repo, "init");
  w(path.join(repo, "openspec", "changes", "change-x", "tasks.md"), tasks(4, 4)); // main works on it, uncommitted

  const wtB = repo + "-b";
  git(repo, "worktree", "add", "-q", "-b", "wb", wtB);

  const x = (await scanOpenSpecAggregated(repo)).activeChanges.find((c) => c.slug === "change-x")!;
  console.log("[2] winner:", x.source?.branch, JSON.stringify(x.taskStats), "— expected main / 4-of-4");
}
```

```
[1] winner: wb {"total":4,"completed":0} — expected wa / 3-of-4
[2] winner: wb {"total":4,"completed":0} — expected main / 4-of-4
```

### The PR already ran into this, in its own test suite

To keep `each change carries its own worktree's defaultSchema` passing, this PR reorders the fixture so `main-change` is added to `main` *after* `wtB` forks, with the comment:

> `// wtB 先分岔，main-change 之後才加到主 repo，兩個 change 才不會因 dedup 而合併成一筆。`

That reordering *is* the bug. Without it, `wtB`'s never-touched inherited copy of `main-change` wins the election — and the schema badge would then be compared against the wrong worktree's `defaultSchema` too. The failure was encountered and worked around as a fixture problem rather than recognised as an algorithm problem.

### Fix direction

The discriminator has to answer "has this copy *advanced beyond what it inherited*", and mtime cannot answer that, because checkout resets it. A git-based candidate filter can:

- A non-main worktree is a candidate for a slug only if it actually diverges there: either its `HEAD` advances `openspec/changes/<slug>/` beyond `main`'s `HEAD`, or it has uncommitted modifications under that path (`git status --porcelain -- <path>` — this covers the "mid-edit, not yet committed" case that `design.md` rightly wanted to keep).
- If no worktree diverges, `main` keeps the change. The unconditional "non-main beats main" rule can then be dropped entirely — divergence subsumes it, and it stops idle worktrees from shadowing `main`.
- mtime survives as a tiebreak *among genuinely diverging copies*, which is the one place it is meaningful.

One trap worth flagging: it is tempting to simplify this to "exclude non-main copies that are byte-identical to `main`'s copy", but the comparison must be against **`main`'s HEAD tree, not `main`'s working tree**. In scenario 2 above, `main` has uncommitted edits, so `wb`'s inherited copy is *not* identical to `main`'s working tree — it would survive the filter and win on its fresh checkout mtime. Compared against `main`'s HEAD it is correctly identified as a pure inherited copy.

The rest of the structure — the shared helper, the graph-side filter point, the spec delta layout — can stay as-is. Only the quantity being compared inside `pickActiveWinners` needs to change.

---

## Non-blocking

**Test coverage doesn't hold the line where it breaks.** The new tests force mtimes with `fs.utimesSync` (2020 vs 2030), which validates the comparison logic but steps around the real-world behaviour that breaks it. Please add a test that creates the worktrees in realistic order — edit in `wa`, *then* create `wb` — with no `utimesSync`, and assert both the winning `source` **and** its `taskStats`. Asserting only `source` is what let the stale-content problem through: that is the part that silently regresses.

**Stale and misplaced docstrings.** `pickActiveWinners` was inserted between `scanOpenSpecAggregated` and its docstring, so the old block ("active changes 聯集不去重並附 source") now sits orphaned above the new function, still describing the old behaviour, while `scanOpenSpecAggregated` has none. `buildGraphDataAggregated`'s docstring still says "active 不去重". `CLAUDE.md` also still describes `scanOpenSpecAggregated` as a non-deduplicating union — the repo workflow requires updating it when archiving.

**`buildGraphDataAggregated` pays for a full scan it doesn't need.** It now runs `scanOpenSpec` per worktree — a `getTimestamps` git subprocess, tasks parsing, artifact counting — purely to obtain a list of active slugs. It already calls `buildGraphData(wt.path)` per worktree, whose change nodes carry both slug and status. Taking `{ wt, slugs: string[] }[]` in `pickActiveWinners` would drop those N full scans entirely. (Separately, `buildGraphData(main.path)` is called twice — once for the spec nodes, once inside the `ordered` loop — pre-existing, but this change adds to the graph path's cost.)

**CHANGELOGs.** Per the two-version-line convention in `CLAUDE.md`, this is an observable behaviour change in the published `@spekjs/core` and needs an entry in `packages/core/CHANGELOG.md`; it is also user-visible across Web / VS Code / IntelliJ, so the root `CHANGELOG.md` needs one too.

---

Happy to look again once the election is based on divergence rather than mtime — the shape of the change is right, and the duplicate-edge / `historyCount` fix is worth landing on its own.

david-lutz:
Good call on the missed use case.  I am pulling this back to draft mode while I consider options for a more complete fix.

dlutz-cc:
I have a new version up that does better git divergence checking, hopefully this solves the problem in a better way.

It looks like I can't move this PR back to "Ready", only user with write access to the this repository can do that.  But is ready for review again.

kewang:
Really nice turnaround — the divergence-based election is the right model, and it clears everything from the last round (comparison against main's HEAD, the graph path reusing the nodes it already builds, the taskStats assertions, docs + CHANGELOGs).

One thing still stands between this and merge. The committed half of `divergedSlugs` uses a two-dot diff:

```
git diff --name-only <mainHead> <wtHead> -- openspec/changes/
```

That's a symmetric endpoint comparison, so it also fires when **main** advances a slug the worktree never touched — the idle worktree then wins the election and the list renders its stale, fork-point copy (wrong taskStats included). Three-dot pins it to the merge-base and counts only the worktree's own advances:

```
git diff --name-only <mainHead>...<wtHead> -- openspec/changes/
```

Heads-up on the guard test named for exactly this (`two-dot vs three-dot — main-side change must not diverge the worktree`): it passes `mainHead` as the `wtHead` argument, which trips the `wtHead !== mainHead` short-circuit and skips the two-dot branch entirely — so it's green without ever exercising the path it's named for. Feed it the worktree's real HEAD and it goes red against the current code, which is what you want it to do.

(Full repro is in the review above.) Fix those two and I'm happy to merge.

david-lutz:
@kewang — I've got the three-dot fix you suggested in progress, but before I push it up I'd like to settle the wider behaviour it implies. One case is genuinely ambiguous, and I'd rather decide it deliberately with you than have it fall out of the implementation.

Setup: a single active change that exists in both `main` and one or more worktrees, all forked from a common point. Whichever copy "wins" is the one the Changes list and Dashboard actually display — including its task-progress numbers — so this decides what people see, not just a label.

Here's how the fix I'm preparing would behave:

| # | Situation | Behaviour | Open? |
|---|---|---|---|
| 1 | The worktree advanced the change; `main` didn't | worktree wins | settled |
| 2 | `main` advanced the change; the worktree only inherited it | `main` wins | settled |
| 3 | **Both** `main` and the worktree advanced it | **worktree wins; `main`'s progress is hidden** | ← open |
| 4 | Two worktrees both advanced it; `main` didn't | the one edited most recently wins | ← open |
| 5 | Nobody advanced it (all just inherited) | `main` wins | settled |

The question is **#3** (and its cousin **#4**): when the same change has been genuinely, differently edited in two places, which copy should the tool show? Right now a worktree beats `main` whenever it has any edits of its own, and between two worktrees the tie goes to whichever was modified most recently.

Three ways it could behave:

- **A — worktree always wins.** Assumes the worktree is where active work happens and `main` is the baseline. Downside: if someone finishes and commits the change on `main` while an old worktree still holds a half-done copy, the tool shows the old worktree's copy and hides `main`'s finished one.
- **B — most-recently-edited copy wins, `main` included.** Symmetric: whoever touched it last wins, regardless of where. Downside: "edited most recently" isn't always "the real one" — a stray touch to a stale copy could make it win.
- **C — show both copies** instead of picking one. Nothing is hidden. Downside: the same change then appears more than once in the list and the graph.

What behaviour would you want for #3 and #4?

kewang:
Thanks for pausing on this rather than letting the implementation decide it — #3 and #4 are exactly the cases worth settling on purpose.

**Decision: go with B for #3, and #4 is already B.** Among copies that have advanced past their merge-base, show the most recently touched one; if none advanced, `main` keeps it. `main` obeys the same rule as every worktree — no special case.

The thing that makes B the right pick rather than a coin-flip: **A is the version with the special case, not B.** A is "a diverging worktree unconditionally beats `main`, even when `main` also genuinely advanced." B just deletes that carve-out and lets `main` compete on the same divergence-then-mtime terms as everyone else. Your #4 handling is already this rule; extending it to `main` in #3 is what makes it one rule instead of two. It also closes the failure mode that made the mtime version a blocker in the first place — an idle half-done worktree can no longer permanently shadow a finished copy, whichever side the finished copy is on.

Worth noting how narrow #3 actually is under three-dot: the common way "someone finishes it on `main`" happens is by merging the worktree's branch back. After that merge, `merge-base(main, wt) == wtHead`, the three-dot diff is empty, and that worktree drops out of the candidate pool on its own — `main` wins without B ever being consulted. The residual #3 that B has to adjudicate is only "both sides edited independently and never merged," which is rare.

**One spec point to pin down so it isn't left to invention** — `main`'s side needs a symmetric definition, since divergence is currently expressed relative to `main`'s HEAD and `main` has no ready-made yardstick:

> `main` enters the candidate pool for slug `S` only when the contest is live (at least one worktree diverges on `S`) **and** `main` has advanced past that worktree's merge-base — i.e. the reverse three-dot `git diff --name-only <wtHead>...<mainHead> -- openspec/changes/S/` lists `S`, **or** `main`'s own working tree has uncommitted edits there (`git status --porcelain -- openspec/changes/S/`). Then mtime tiebreaks among all candidates.

When no worktree diverges, none of this fires and `main` keeps the change trivially. This is a dozen lines on top of the shape `divergence.ts` already has plus one extra git call per competing worktree — negligible.

---

**On the "show all copies" idea (C) and a global A/B/C toggle: I'd rather not.** Two different reasons for the two halves:

- A vs B isn't a choice a user can act on — in almost every repo state the two produce byte-identical output, so a switch between them is a control with no visible effect. That's not a view mode like the artifact sort (same items, reordered, nothing hidden); it's picking an algorithm.
- C as a mode brings back duplicate change nodes and the inflated `historyCount` in the graph — the exact bug this PR exists to remove. I don't want to ship the bug as a toggle position.

The legitimate worry underneath C is real, but it only bites in #3/#4: a genuine fork where both copies are live, and picking one hides the other's progress. **I'd handle that by surfacing the losers rather than switching the winner** — a small badge on the winning row ("also edited in `wa`") that links into the other copy via the `?wt=` param the detail endpoint already takes. Nothing is hidden, and because the trigger is "diverged-but-lost" (not "some other worktree also holds a copy" — inheritance means every worktree holds nearly every change, which would light the badge on every row), it stays rare and every appearance is a real signal.

**Please keep that badge out of this PR, though.** This one's scope is the election fix, and B + three-dot is a small delta on what you already have. The badge is a new `ChangeInfo` field plus UI, and bolting it on would drag a PR that's already been around three rounds through another one. The diverged-losers list is already sitting in the `diverging` array inside `pickActiveWinners` and just being dropped, so it's cheap to expose in a follow-up — I'll open a separate change for it.

So for this PR: the three-dot fix you have in progress, `main` folded into the candidate pool per the definition above, and the guard test fed the worktree's real HEAD so it actually exercises the two-dot path. Push it up and I'll take another look.

kewang:
Merged as a146741. Thanks for staying with this — the result is a better fix than a single pass would have produced, and a few things are worth calling out.

**You paused on #3/#4 instead of letting the implementation decide them.** This is the one I'd highlight. It would have been easy to ship whatever fell out of the code and let the ambiguity resurface later as a bug report. Asking first is what turned a coin-flip into a rule that can be stated in one sentence — and stating it is what made it obvious that A, not B, was the version carrying the special case.

**You treated the mtime→divergence correction as a model problem, not a test problem.** The first revision had already worked around the bug inside its own fixture — reordering `main-change` so dedup wouldn't merge it. Recognising that the reordering *was* the bug, and rebuilding on git divergence instead of patching around it, is the hard call in this whole PR.

**A couple of things went past what was asked.** `listActiveChangeSlugs` makes the list and graph election inputs identical *by construction* rather than by inspection — that closes the drift the review only asked you to fix once. Hoisting main's `git status` out of the per-contender loop wasn't requested either. Both are right.

I checked that the guard test bites this time: reverting three-dot back to two-dot turns it red, along with `divergedSlugs: uncommitted edit diverges; a slug main later advanced does not`. That's the difference from last round, where the test was green without ever reaching the path it was named for.

Two maintainer commits went on top before the squash, so you're not surprised by them:

- **27d1ae7** restores `scanner.ts`'s CRLF line endings. The branch had flipped the whole file to LF, which rendered a +148/-16 change as a +666/-534 rewrite and would have reattributed all 666 lines in `git blame`. This one isn't on you — the repo is mixed-line-ending with no `.gitattributes`, so an editor on a different platform does this silently and the diff gives no hint. Normalising it repo-wide is worth doing, but as its own change.
- **8671495** renames the divergence helpers' parameters to `(repoPath, aheadHead, baseHead)`. They're called in both directions, so `wtHead`/`mainHead` read as their exact opposites at the main-side call site — which is the same confusion that produced the vacuous guard test last round. Internal helpers, positional args, no behaviour change.

Follow-up as promised: the diverged-losers are still being dropped on the floor inside `pickActiveWinners`, so the "also edited in `wa`" badge is cheap from here. I'll open that separately.

PR #19, opened 2026-07-16
fix(core): key schema-order cache by schema, not by change (#15)

### LINKED ISSUE #15: Change detail takes ~1.25s: the openspec CLI is spawned once per change to compute a per-schema value
Opening any active change's detail page costs **~1.25 seconds**, essentially all of it spent spawning the `openspec` CLI to compute `schemaOrder` — a value that is only used if the user switches the artifact sort to "Schema order" (which is *not* the default; the default is "Last modified").

## Measured

Against this repo (71 changes), with the `openspec` CLI installed:

```
1. scanOpenSpec (changes list, 71 changes)
     93.9 ms

2. readChange (one active change's detail)
   a) default (spawns `openspec status --change <slug> --json`)
      cold : 1249.4 ms
      warm : 0.7 ms   (30s TTL cache hit, same slug)
   b) with a no-op order provider (never touches the CLI)
           : 0.5 ms

   → CLI spawn cost = 1248.8 ms   (99.96% of the request)
   → pure file I/O  = 0.5 ms
```

The CLI is Node-based, so almost all of that is process startup, and it is paid **per change**:

- open change A → 1.25 s
- open change B → **another 1.25 s** (different slug ⇒ cache miss)
- re-open A after 30 s → **another 1.25 s** (TTL expired)

## Root cause

`schema-order.ts` keys its cache by the **change**:

```ts
const cacheKey = `${repoRoot}::${slug}`;   // schema-order.ts:103
```

But the thing the CLI returns — `actionContext.planningArtifacts` + `artifactPaths` — is a property of the **schema**, not of the change. Every `spec-driven` change in a repo gets the identical answer. We are spawning one process per change to ask the same question over and over.

(`resolveSchemaOrder(refs, knownIds)` *is* per-change, but it's pure and costs nothing — it just maps the schema's refs onto that change's discovered artifact ids.)

The Kotlin core has the same key (`SchemaOrder.kt:108`), so the IntelliJ plugin pays the same cost.

## Fix

**Key the cache by schema instead of by change** — `${repoRoot}::${schema}`. The CLI invocation still needs *a* slug to run against, but the result is reusable for every change sharing that schema.

- First `spec-driven` change opened → 1.25 s (unavoidable, one CLI call)
- Every subsequent `spec-driven` change → ~0.5 ms
- A repo with 71 changes goes from *up to 71 spawns* to **one spawn per distinct schema**

Cheap, local to `schema-order.ts` (+ the Kotlin mirror), no API or UI change. Fall back to keying by slug when a change's schema is `null`/unknown.

### Possible follow-on (larger)

Make it genuinely lazy — only fetch `schemaOrder` when the user actually selects "Schema order". That removes the cost entirely for everyone who never touches the sort control, but needs a new endpoint/adapter method across the three frontends. The cache-key fix above takes most of the pain for a fraction of the effort, so it's worth doing first and measuring again.

## Context

This was flagged qualitatively during the #6 review ("the eager-spawn cost is real and worth a conscious decision rather than a side effect") but shipped without a measurement. With the number in hand, 1.25 s on every change-detail open is user-perceptible and worth fixing.

nthansen:
## Problem

Opening any active change's detail page costs **~1.25s** (measured against this repo), essentially all of it spent spawning the `openspec` CLI to compute `schemaOrder` — a value only used when the artifact sort is set to **Schema order** (not the default, which is Last modified).

Worse, it was paid **per change**: the cache was keyed by change slug (`schema-order.ts`, `SchemaOrder.kt`), but the CLI's `actionContext.planningArtifacts` + `artifactPaths` depend only on the change's **schema**. Every `spec-driven` change in a repo spawned a fresh process to compute the identical answer — up to N spawns for N changes.

## Fix

Key the cache by `${repoRoot}::${schema}` instead of by slug, so all changes sharing a schema reuse **one** CLI spawn (one per distinct schema per repo, within the existing 30s TTL). The per-change `resolveSchemaOrder` mapping still runs on every read, so each change's `schemaOrder` is unchanged — a change's own artifact set still determines its ordering; only the schema-level authority is shared.

- A change with **no schema** has no authoritative order, so the caller skips the provider entirely (no spawn) rather than keying by slug. The provider's `schema` argument is therefore non-null (compile-enforced on the Kotlin side).
- An **empty slug** — which resolves to the `changes/` directory itself and would otherwise poison the schema bucket with a null result — is rejected up front.
- Mirrored in the Kotlin core (the IntelliJ plugin paid the same per-change cost).

No API, adapter, or UI change; the `ChangeDetail.schemaOrder` contract is untouched.

## Measured impact

| | Before | After |
|---|---|---|
| First `spec-driven` change opened | ~1.1s | ~1.1s (one unavoidable CLI call) |
| Every subsequent change sharing that schema | ~1.1s | **~1.5ms** (cache hit, no spawn) |

## Testing

- **`@spekjs/core`**: 125 unit tests pass. New tests assert the schema is threaded to the provider, that a schemaless / empty-string-schema / empty-slug change never calls the provider, and that two shared-schema changes each map the cached order onto their own discovered artifacts.
- **IntelliJ**: full instrumented `./gradlew clean test` green (`ChangeReaderTest`, `SchemaOrderTest`). The order provider is injectable so the guard is asserted deterministically without invoking the real CLI.
- **Mutation testing** on the new guard logic — no surviving mutants beyond pre-existing ones in untouched code.
- **End-to-end (Playwright)**: switching to "Schema order" reorders the artifact tabs to the schema's authoritative sequence; a fresh/empty change renders gracefully (no tabs, no crash); zero console errors.

Fixes #15.


kewang (changes requested):
Thanks for this — the analysis is excellent and the core of the change is right.

I verified the central premise against the real CLI before looking at the rest: two changes in
one repo with different artifact sets return identical `actionContext.planningArtifacts`
(`["proposal","design","specs","tasks"]`) and identical `artifactPaths` keys. So re-keying the
cache by schema is correct and the win is real. 125 core tests pass here, type-check is clean,
and it merges cleanly onto master. The empty-slug guard is a genuine catch.

## Blocker: the "no schema → skip the CLI" guard is a regression

The justification in the spec —

> When a change has no schema (its resolved schema is null or empty), there is no authoritative
> order to obtain

— doesn't hold. The CLI resolves a default schema on its own. Against a repo with no
`openspec/config.yaml`, for a change that declares no schema:

    $ openspec status --change a --json
    {
      "changeName": "a",
      "schemaName": "spec-driven",
      "planningHome": { "kind": "repo", ..., "defaultSchema": "spec-driven" },
      "artifactPaths": { "proposal": {…}, "specs": {…}, "design": {…}, "tasks": {…} }

A null schema means *spek couldn't resolve the name locally* — not that no authoritative order
exists. So the guard drops an order that is being delivered today:

    repo: no openspec/config.yaml, change with no .openspec.yaml
    master:  schemaOrder = ["proposal","design","tasks"]
    this PR: schemaOrder = undefined      ← Schema order silently falls back

This isn't an exotic setup: `openspec init` doesn't create `config.yaml` at all — it only creates
`openspec/specs`, `openspec/changes`, and `openspec/changes/archive`. It escapes notice here only
because changes usually carry their own `.openspec.yaml` (74 of 75 in this repo) and spek's own
`config.yaml` happens to set `schema:`. The tests added in this PR pin the new behavior as
intended, so nothing flags it.

## The fix is the alternative you rejected

From design.md:

> *共用一個 null 桶*：兩個無 schema 的 change 會互相污染。否決。

They don't pollute each other. Every schema-less change in one repo resolves to the same
`planningHome.defaultSchema` — it's a repo-level property, visible in the JSON above. So one
shared bucket is a *correct* share, not a collision:

- `SchemaOrderProvider`'s `schema` param becomes `string | null`
- key: `${repoRoot}::${schema ?? "\0default"}`
- drop the `&& schema` condition in `readChange` / `ChangeReader.read` — the empty-slug guard
  stands on its own and should stay

That keeps the full perf win (one spawn per distinct schema, plus one for the repo default) with
no behavior change. A change that explicitly declares `spec-driven` and one that inherits it would
spawn twice rather than once — a fine price for not silently dropping the order. Worth pinning
with a test: two schema-less changes in one repo share a single spawn.

## Minor

**Shared buckets amplify transient failures.** If the CLI times out for one change, `null` is now
cached for every change sharing that schema for the 30s TTL, where before only that change
degraded. Bounded and it degrades gracefully, so not blocking — but it's a real consequence of the
re-key, worth a line under Risks in design.md.

The design doc, the spec text, tasks.md, and the CLAUDE.md paragraph all carry the same premise,
so they'd need to follow the code. Happy to take it on if you'd rather not — but the D1 reasoning
is yours to rewrite, and you have the IntelliJ toolchain to re-run the gradle / mutation /
Playwright verification your tasks.md records, which I can't reproduce here.

nthansen:
Reproduced it: in a repo with no `config.yaml`, `openspec status --change a --json` returns `schemaName: spec-driven` + full `planningArtifacts` for a change with no `.openspec.yaml`, and `readChange` gave `undefined` where master gives `["proposal","design","tasks"]`. The `&& schema` guard was dropping an order the CLI delivers, and the tests I added pinned that behavior — so nothing flagged it.

Fixed in 3820e77:

- Dropped the `&& schema` guard — every active change consults the provider again. `schema` is `string | null` (and `String?` in Kotlin).
- Schema-less changes share **one repo-level default bucket** (`${repoRoot}::\0default`) rather than keying by slug — they all resolve to the same `planningHome.defaultSchema`, so it's a correct share (one spawn per repo default), as you noted. Pinned: two schema-less changes in one repo hit the same bucket / single spawn (null and `""` fold into it).
- Kept the empty-slug guard.
- The regression tests are inverted: a schema-less change now asserts it **still** consults the provider and gets the default order (`["proposal","design","tasks"]`), matching master; empty-string schema likewise.
- Minor: added the shared-bucket transient-failure consequence to Risks in `design.md` — a cached CLI timeout now applies `null` to every change sharing that schema for the 30s TTL, where before only the one change degraded. Bounded, degrades gracefully.
- `spec` text, `design.md` (D1 + the rejected-alternative reasoning), `tasks.md`, and the `CLAUDE.md` paragraph all follow the code now.

Full validation re-run (the gradle / mutation / Playwright list you flagged):

- **127 core tests**, type-check — green.
- **Instrumented IntelliJ `clean test`** (Temurin 17) — green (ChangeReaderTest 8, SchemaOrderTest 14, no `w:` warnings).
- **Mutation** on the `readChange` guard range — no new survivors (the lone survivor is the pre-existing `existsSync` metadata line, which this change doesn't touch).
- **Playwright**, driving a no-config repo with a schema-less change: "Schema order" delivers the CLI default (Proposal → Design → Tasks), distinct from Last-modified (Tasks → Design → Proposal), with no fallback message and 0 console errors.
- **End-to-end**: `readChange` on that repo returns `["proposal","design","tasks"]` (= master) instead of `undefined`.

kewang (approved):
Verified independently — thanks for the thorough turnaround.

- **Regression gone.** Same repro (no `config.yaml`, schema-less change): `schemaOrder` is back to
  `["proposal","design","tasks"]`, matching master.
- **The perf win survived the fix, on both paths** — declared schema 1918ms -> 0.9ms, and
  schema-less 6555ms -> 8.7ms via the shared default bucket. Schema-less changes now get both the
  correct order *and* a single spawn, which is better than master (one spawn per change) and better
  than the first revision (no order at all).
- **Tests re-run here**, not taken on trust: 127 core + type-check green; `./gradlew clean test` on
  JDK 17 gives 80 tests / 0 failures, `ChangeReaderTest` 8 and `SchemaOrderTest` 14 — exactly your
  numbers.
- Spec, design.md, tasks.md and the CLAUDE.md paragraph all track the code now.

Two things you got better than my suggestion, worth noting: `||` rather than `??` (so `""` folds
into the default bucket instead of leaving a degenerate `${repoRoot}::` key), and `"default"`
in Kotlin — `\0` isn't a valid Kotlin escape, so copying my draft verbatim would not have compiled.

One byte to clean up, and it's my fault for suggesting the sentinel: `tasks.md` line 17 (task 3.2)
contains a **literal NUL** rather than the two characters `\0` — task 1.2 has it right, so the
escape got interpreted somewhere in transit. Git now treats that file as binary
(`Bin 3408 -> 3793`, no diff/blame/textual merge), and it would ride into `docs/demo.html` on the
next demo build. I managed to do exactly the same thing while drafting this comment, so the trap is
real. Not worth another round trip — I'll fix it on merge, along with the CHANGELOG entries, which
are a convention we've never actually written down for contributors (that's on us, not you).

Merging. Thank you — the analysis, the reproduction, and the re-verification were all excellent.

PR #21, opened 2026-07-18
fix(web): keep loose-list markers on the same line as their content

nthansen:
Fixes the one real bug: in a **loose** list (items separated by blank lines, so each item's content is wrapped in a block `<p>`), the bullet/number was pushed onto its own line *above* the text — an `list-inside` marker can't share a line box with a block element.

**Fix:** switch `ul`/`ol` from `list-inside` to `list-outside` with a `pl-6` marker gutter. The marker sits in the gutter, inline with the first line; wrapped lines hang-indent under the text. One component, effectively a two-class change.

Markers, numbering, checkboxes, and task-list rendering are otherwise unchanged.

---

**Scope note (why this was re-opened much smaller):**

An earlier revision of this branch went further — it suppressed the list marker on task items (GitHub-style) and reworked the checkbox. That was out of scope, and it had a real downside: suppressing the marker on an *ordered* task item hides its **number**, so a mixed list like

```
1. Step to do something
2. [ ] Checkbox step
3. Do this action
```

rendered as `1. → (no 2) → 3.`, losing the sequence. This PR is deliberately reduced to just the loose-list line-break fix — nothing about markers or numbering changes.


nthansen:
This change impacted more than I would have liked so I'm going to close for now until I think it's ready for review

PR #22, opened 2026-07-18
fix(web): keep loose-list markers on the same line as their content

nthansen:
In a **loose** list — items separated by blank lines, so each item's content is wrapped in a block `<p>` — the bullet/number is pushed onto its own line *above* the text. A `list-inside` marker can't share a line box with a block element.

## Repro

Any loose list. Standard CommonMark — no custom schema, no checkboxes required:

```markdown
- First item

- Second item

- Third item
```

loose-list-marker-before-after

I hit this viewing a change authored with the `superpowers-bridge` schema, but it isn't schema-specific — any loose list reproduces it, including a stock `proposal.md`.

## Fix

Switch `ul`/`ol` from `list-inside` to `list-outside` with a `pl-6` marker gutter. The marker sits in the gutter, inline with the first line; wrapped lines hang-indent under the text. One component, effectively a two-class change.

Two honest notes:

- **All lists** — not just loose ones — now use a hanging indent, with the marker in a 24px gutter instead of sitting flush inline. That's standard list rendering and is inherent to `list-outside`; loose and tight can't be distinguished at the `ul` level, so it can't be scoped to loose lists only.
- A flat ordered list of **100+ items** would have 3-digit markers slightly wider than the 24px gutter. Nothing clips (no `overflow-hidden` ancestor; the page's own padding absorbs it), so it's cosmetic and effectively unreachable here.

---

**Scope note:** this supersedes #21, which was closed. That earlier revision also suppressed the list marker on task items (GitHub-style) and reworked the checkbox. That was out of scope, and it had a real downside: suppressing the marker on an *ordered* task item hides its **number**, so a mixed list like

```
1. Step to do something
2. [ ] Checkbox step
3. Do this action
```

rendered as `1. → (no 2) → 3.`, losing the sequence. This PR is deliberately reduced to just the loose-list line-break fix.

### DIFF
--- a/packages/web/src/components/MarkdownRenderer.tsx
+++ b/packages/web/src/components/MarkdownRenderer.tsx
@@ -225,12 +225,14 @@ export function MarkdownRenderer({ content, specTopics, idPrefix }: MarkdownRend
               </td>
             );
           },
-          // 列表
+          // 列表：用 list-outside + pl-6，marker 落在左內距溝槽並與內容首行同行。
+          // list-inside 會讓 loose list（項目間有空行、內容被包成 block <p>）的 marker 被
+          // 擠到自己一行，因為 inline 的 marker 無法與 block <p> 共用 line box。
           ul({ children }) {
-            return <ul className="list-disc list-inside mb-4 space-y-1">{children}</ul>;
+            return <ul className="list-disc list-outside pl-6 mb-4 space-y-1">{children}</ul>;
           },
           ol({ children }) {
-            return <ol className="list-decimal list-inside mb-4 space-y-1">{children}</ol>;
+            return <ol className="list-decimal list-outside pl-6 mb-4 space-y-1">{children}</ol>;
           },
           // 分隔線
           hr() {


kewang:
Thanks for this — and especially for reducing it from #21 down to just the loose-list fix. That was the right call.

I pulled the branch and verified the fix holds up. Measuring the rendered geometry with the compiled CSS:

| Case | `list-inside` (before) | `list-outside pl-6` (this PR) |
|---|---|---|
| Loose bullet list | li height **1.93 lines** | li height **0.96 lines** |
| Tight bullet list (control) | 1 line | 1 line |

The 1.93 is the marker occupying its own line — exactly as you described. Your root-cause analysis is correct, and the tight-list control confirms the trigger is looseness, not checkboxes. `npm run build` and `npm test` (36/36) both pass on the branch.

## One thing I'd like to adjust: the `ol` gutter

While testing I found the 24px gutter is a little tight for numbers. Measured marker widths:

| Marker | Width | Clearance in a 24px gutter |
|---|---|---|
| `9.` | 13.3px | 10.7px |
| `10.`–`99.` | 22.2px | **1.8px** |
| `100.` | 31.1px | **overflows by 7.1px** |

So it isn't only the 100+ case you flagged — **two-digit markers already sit flush against the list's left edge** with under 2px of breathing room. And since `list-inside` keeps the marker inside the box as inline content, this particular effect is new with `list-outside`.

It's cosmetic only (nothing clips — no `overflow-hidden` ancestor), but a numbered list of 10–20 steps is common in a proposal, so it's worth getting right.

The fix that keeps everything you gained: **leave `ul` at `pl-6`, bump only `ol` to `pl-8`.**

```diff
- <ol className="list-decimal list-outside pl-6 mb-4 space-y-1">
+ <ol className="list-decimal list-outside pl-8 mb-4 space-y-1">
```

Bullets are a fixed ~7px, so 24px is already ample and widening them would just add indentation for nothing. Numbers are variable-width, so only `ol` needs the wider gutter. Screenshot comparing all three (all rendered with the branch's compiled CSS):

pr22-recommended

## Plan

I'm going to **merge this as-is** and push the `ol` tweak as a follow-up, along with two housekeeping items on my side: translating the new comments to English (this repo keeps everything committed in English — no way you'd have known, the surrounding comments are still Chinese), and rebuilding `docs/demo.html` + the VS Code webview bundle, which are committed build artifacts that still carry the old CSS.

Thanks again — good catch and a clean, well-scoped fix.

PR #23, opened 2026-07-19
feat: experimental jj (Jujutsu) workspace aggregation

DannyGoodall:
## Summary

Adds **experimental, opt-in** aggregation of OpenSpec changes from **Jujutsu (jj) workspaces**, alongside the existing git-worktree aggregation. It's **off by default**, and when disabled — or when `jj` isn't installed — behaviour is byte-for-byte identical to today. Your 1.8.1 divergence de-dup (`divergence.ts` / `pickActiveWinners`) is **untouched**: jj rides in as a parallel path behind the existing `WorktreeInfo` model. Rebased onto `v1.8.3` (LF-normalized).

The one-paragraph idea: spek already builds its aggregated view from an internal list of "working directories". A jj workspace becomes another entry in that list, tagged `vcs: "jj"`. jj is invisible to git, so it's discovered by asking the `jj` CLI directly; and because jj duplicates changes differently from git worktrees, the de-dup step picks a winner *per source* — git worktrees keep going through your history-based election, jj workspaces are de-duped by comparing file contents.

## Updates from review (thanks @kewang)

- **Fixed the Web jj-toggle no-op** — `FetchAdapter` only appended the `jj` param when *false*, but the server defaults jj off, so the toggle-on case sent nothing. Corrected to append `jj=true` when on, and added a unit test on the query polarity so it can't silently regress.
- **Collapsed the two checkboxes into a tri-state control** — `aggregate` and `jj` weren't really independent (jj is only meaningful *inside* aggregation), which allowed a contradictory "aggregate off + jj on" state. The Web UI now has one control: **Current dir / Worktrees / Worktrees + jj** — the invalid state is unrepresentable. (VS Code stays on the `spek.aggregateJjWorkspaces` setting, which the handler treats as the source of truth.)
- **Rebased onto `v1.8.3` and normalized to LF**, and re-did the CLAUDE.md jj notes in **English** to match the new English-canonical doc.

## What changed & where

**Core — `packages/core/src/`**
- **`jj-workspaces.ts`** *(new)* — all jj-specific I/O: `listJjWorkspaces()` (`jj workspace list`), `parseJjWorkspaceList()` (pure), `jjCurrentChangeSlugs()` (`jj diff --ignore-working-copy -r @`, read-only).
- **`worktrees.ts`** — new `listWorkspaces(dir, { includeJj })` merges git worktrees + jj workspaces and de-dups the colocated main dir by path (git entry wins, so the branch is kept). `WorktreeInfo` gains `vcs: "git" | "jj"`.
- **`scanner.ts`** — `scanOpenSpecAggregated` and `buildGraphDataAggregated` now gather via `listWorkspaces` and split de-dup by source; adds `changeContentFingerprint()`. **`divergence.ts` / `pickActiveWinners` unchanged.**
- **`types.ts`** — `WorktreeInfo.vcs`; `ChangeInfo.isCurrent?` / `ChangeInfo.conflictsWith?`.

**Web — `packages/web/`** — `server/routes/openspec.ts` (`jj` query param → `includeJj`; `/watch` enumerates via `listWorkspaces`), the API adapters (`FetchAdapter.ts` param fix + `aggQuery` extracted for testing), `pages/ChangeList.tsx` + `utils/{jjWorkspacePref,aggregationLevel}.ts` (the tri-state aggregation control + jj source / editing / conflict markers).

**VS Code — `packages/vscode/`** — `package.json` setting `spek.aggregateJjWorkspaces` (boolean, `default: false`, `tags: ["experimental"]`); `handler.ts` / `panel.ts` / `tree-provider.ts` read the setting (it's the single source of truth in VS Code).

**Specs — `openspec/`** — extends `worktree-aggregation`, adds `jj-workspace-aggregation`, touches `graph-view` / `live-reload` / `openspec-api` / `vscode-sidebar`. Change record archived under `openspec/changes/archive/2026-07-16-aggregate-jj-workspaces/`.

**Docs — `CLAUDE.md`** — adds a jj section so anyone developing the extension further has the context in one place: the VCS-dispatched aggregation, `listJjWorkspaces` / `listWorkspaces` / `jjCurrentChangeSlugs`, the API `jj` param, and the `spek.aggregateJjWorkspaces` setting — all explicitly flagged **EXPERIMENTAL / off by default**, matching the feature's opt-in framing.

## How it fits the 1.8.1 de-dup (the one non-trivial bit)

Your de-dup picks a winner by comparing git history — which copy of a change has advanced past the main worktree. That can't be applied to jj, for two reasons: a jj working copy is identified by a **change-id, not a git commit** (so the history comparison never runs — and the code treats "git failed" as "didn't diverge", which would silently drop the workspace), and every jj workspace **shares one history and materialises the full trunk** (so the same change appears in *every* workspace, identical). So `scanOpenSpecAggregated` splits the input:

- **git worktrees** → your history-based election, verbatim (advanced-past-merge-base wins, mtime tiebreak).
- **jj workspaces** → de-dup by **content fingerprint** (a hash of the change dir's relative paths + bytes): identical copies collapse to one; a divergent copy (a workspace editing that change) is kept and flagged `conflictsWith` its base; the `@`-edited one is flagged `isCurrent`.

The two are combined by slug afterward — jj entries are never handed to `pickActiveWinners`. Same split in `buildGraphDataAggregated`.

## Degradation / failure modes

- **Plain git repo / no `jj` binary** — `listJjWorkspaces` returns `[]`, a `repoUsesJj` guard skips the whole jj path; zero extra I/O (fingerprints only computed when jj workspaces exist).
- **Colocated repo, git-only user** — the main dir (git worktree *and* jj `default` workspace) folds to a single entry; nothing extra shows without additional jj workspaces.
- **Pure-jj repo** (no git backing) — handled; aggregation runs entirely through the jj content path.
- **Opt-in** — even with jj installed, nothing changes until `spek.aggregateJjWorkspaces` / the web toggle is enabled.

## Testing

- `@spekjs/core`: **161 / 161** unit tests (0 skipped), including a git↔jj coexistence test asserting the two de-dup paths don't interfere.
- `@spekjs/web`: **42 / 42** — adds `aggQuery` (param-polarity regression guard) and `aggregationLevel` (tri-state ↔ prefs) tests.
- `openspec validate --all`: 45 / 45.
- Live colocated smoke: a jj-workspace change surfaces with `source.vcs: "jj"` + `isCurrent`; the colocated main de-dups to the git entry.
- `type-check`, `build:core`, `build:web`, `build:vscode` all green.

## Notes for landing

- Happy to keep it **experimental / feature-flagged** as discussed — the setting already defaults off.
- Single commit (the small VS Code webview fix that makes the setting authoritative is folded in).
- **`CLAUDE.md`** now carries the jj architecture notes (explicitly flagged **EXPERIMENTAL**), so anyone extending the extension has the jj context alongside the existing docs. Happy to adjust the wording/voice to match your style.
- **IntelliJ + Demo are untouched** — the IntelliJ plugin's Kotlin core has no worktree-aggregation foundation yet, so jj parity there would be a separate, larger effort.


kewang:
Hi Dan — heads up before I land this: I'll do the merge on my side, so no action needed from you on the conflicts. But I want to flag *why* they happen, for your future PRs.

On 2026-07-18 (after your branch point) master normalized the whole repo to LF and added a `.gitattributes`. Your branch predates that and is still CRLF, so a plain merge marks **every line** of `scanner.ts` / `types.ts` / `index.ts` as changed — it looks like a huge core conflict, but it's ~90% line-ending noise. The actual jj change in `scanner.ts` is only ~170 lines and applies cleanly once endings match (`git merge -X renormalize` makes the core conflicts disappear entirely).

For anything you send in future, it'd help to renormalize your branch to LF first:

```bash
git add --renormalize .
git commit -m "chore: normalize line endings to LF"
```

(with the repo's `.gitattributes` in place). That keeps your diffs clean and avoids the whole-file churn.

I'll take the jj changes from here — will ping you if the merge surfaces anything real. Thanks again for the solid write-up and tests.

Cheers, kewang

kewang:
Hi Dan — I took this end-to-end against a real colocated git+jj repo (jj 0.43: one added workspace with a jj-only change, another with a divergent copy of a shared change). The core is genuinely solid — all 161 core tests pass with jj installed (the integration tests run, not skipped), and the dedup election + content-fingerprint paths behave exactly as designed: `jj:<name>` source, `editing` (isCurrent), and `conflicts with main` all surface correctly. Two things came out of the run — one bug, one design call.

**1. Bug (Web): the jj toggle never actually enables jj**

`FetchAdapter.agg()` only appends a jj param when it's *false*:

```ts
(includeJj === false ? "&jj=false" : "")
```

But the server defaults jj *off* (`req.query.jj === "true"`), so when the toggle is **on** the adapter sends nothing and the server reads jj as off. Net effect: the Web "Include jj workspaces" toggle is a no-op — I confirmed the outgoing request stays `/changes?dir=…` with no `jj` param even with the box checked. It affects `/changes`, `/overview`, and `/graph` (all route through `agg()`). VS Code is unaffected — it reads the `spek.aggregateJjWorkspaces` setting in the handler, not this adapter.

The polarity looks mirrored from `aggregate`, which defaults *on* (so omit-when-true is correct there). Since the server default for jj is off, the one-liner is:

```ts
(includeJj === true ? "&jj=true" : "")
```

Verified: with that change the request carries `&jj=true`, the toggle shows, and the page renders the full aggregated view with the jj rows + markers.

**2. Design: `aggregate` and `jj` are two independent toggles, but jj is really a sub-option of aggregate**

The way I read it: jj workspaces are just another kind of working directory — invisible to `git worktree list`, so they're discovered via `jj workspace list` and folded into the same `WorktreeInfo` set. That means `includeJj` only has meaning *inside* aggregation: with `aggregate` off, `scanOpenSpecAggregated` short-circuits to a single-dir scan regardless of jj.

So two independent checkboxes allow a contradictory combo — **aggregate off + jj on**: the jj box is checked but does nothing, while the label still reads "Aggregate 4 worktrees" (it counts jj workspaces) even though only the current dir's changes are shown. The count also flips 2↔4 as you toggle jj, which is a tell that the two aren't really independent.

Rather than patch the visibility rules, I think the clean model is a single tri-state control for "what to aggregate":

- **Off** — current directory only
- **Git worktrees**
- **Git worktrees + jj workspaces**

That folds the dependency into the structure — there's no way to land in an invalid state. Curious whether you see it the same way.

Neither is urgent, and both sit on top of the earlier CRLF/renormalize note. Thanks again — the underlying implementation is clean; these are wiring/UX rather than logic.

Cheers, kewang

DannyGoodall (commented):
In addition to the changes discussed, the CLAUDE.md file was also updated to provide agents with the addition JJ context.

DannyGoodall:
Sorry @kewang, I missed your previous comment when I posted my last reply. 

1. Web toggle bug —  when I flipped jj to default-off I updated the server, the VS Code handler and the localStorage default, but missed the adapter (and its now-stale comment). Fixed to append jj=true when on, with a small unit test on the query-building so it can't silently regress.

2. Aggregate/jj as one control — completely agree - jj really is a sub-level of aggregate, and the two independent checkboxes did allow the contradictory "aggregate off + jj on". I've replaced them (Web) with your tri-state — Current dir / Worktrees / Worktrees + jj — so the invalid state is unrepresentable, and dropped the misleading count. Two notes: it's a Web-UI change (VS Code stays on the spek.aggregateJjWorkspaces setting, which the handler treats as the source of truth), and in a pure-jj repo the middle level naturally degrades to current-dir.

3. CRLF — thanks for the write-up, and sorry for the churn. Rather than hand you a -X renormalize merge, I've rebased onto v1.8.3 and renormalized to LF, so it should apply straight now. While doing that I saw you'd moved CLAUDE.md to English-canonical — I've redone the jj notes I'd added there in English to match, rather than leave a Traditional-Chinese block. I'll keep future branches LF from the start per your note.

Do you want the Traditional Chinese OpenSpec text to be converted too? I could do that as a follow-up.

All pushed to the same PR (still a draft while you have eyes on it). Thanks again for the thorough test pass!

Cheers, Dan

kewang:
Thanks Dan — this is all verified on my side, and it's good to merge.

- **Adapter fix**: correct, and I really like that you added the query-building test as a regression guard. Ran it end-to-end — selecting "Worktrees + jj" now sends `jj=true` and the jj rows + `editing` / `conflicts with main` markers show up.
- **Tri-state**: exactly right. The invalid `aggregate off + jj on` state is now unrepresentable, and the round-trip test locks it in.
- **Rebase / LF**: clean rebase onto 1.8.3, LF throughout, merges without conflict. And thanks for redoing the CLAUDE.md notes in English to match — appreciated.

Tests are green here too: core 161/161 (jj integration tests actually running) and web 42/42 including your two new ones.

I'll take it from here for the merge and handle the remaining housekeeping on our side (the product CHANGELOG lines, etc.) — nothing more needed from you. And no need to convert the Traditional-Chinese OpenSpec text; we'll handle that ourselves as a separate follow-up if we go that way.

Thanks again for the fast turnaround and the thorough tests — genuinely nice work.

Cheers, kewang

Issue #16, opened 2026-07-12
No CI: nothing runs tests or type-checks; type-check only covers packages/web


kewang:
There is no CI that runs tests or type-checks. The three workflows in `.github/workflows/` are all publish/deploy:

| workflow | trigger | what it does |
|---|---|---|
| `vscode-publish.yml` | push **tags** | publish to VS Code Marketplace |
| `intellij-publish.yml` | push **tags** | publish to JetBrains Marketplace |
| `pages.yml` | push | deploy `docs/` verbatim |

Nothing runs on `pull_request`. Nothing runs `npm test`, `tsc`, or `./gradlew test`. Every quality gate is currently a thing a human remembers to run locally.

## This already bit us

In #13, `ChangeInfo` gained a **required** `defaultSchema` field. That left `packages/ui/src/timeline/__tests__/grouping.test.ts` type-invalid:

```
grouping.test.ts(7,3): error TS2741: Property 'defaultSchema' is missing
                       ... but required in type 'ChangeInfo'.
```

`npm test -w @spekjs/ui` reported **22/22 passing** the whole time. It surfaced only because I ran an ad-hoc `tsc` against that one file by hand during review. Two independent holes had to line up for that:

1. **`npm run type-check` only covers `packages/web`** — it is literally `npm run type-check -w @spekjs/web`. `core`, `ui` and `vscode` have no `type-check` script at all.
2. **`packages/ui/tsconfig.json` excludes its own tests** — `"exclude": ["src/**/__tests__/**"]` — and the suite runs through `tsx`, which strips types without checking them. So a type-invalid test file runs green.

`@spekjs/core` is published to npm, so the class of bug that slipped through here is exactly the class that reaches downstream consumers.

## Current state

```
root scripts:
  test:       npm test -w @spekjs/core && npm test -w @spekjs/ui && npm test -w @spekjs/web   ✅ covers 3
  type-check: npm run type-check -w @spekjs/web                                                ❌ covers 1
  lint:       (none)

per-workspace:
  core    test ✅   type-check ❌
  ui      test ✅   type-check ❌
  web     test ✅   type-check ✅
  vscode  test ❌   type-check ❌

Kotlin (packages/intellij): ./gradlew test exists, never runs in CI
```

## Proposed fix

**1. Add `.github/workflows/ci.yml`** on `pull_request` + `push: [master]`:

- `npm ci`
[...]

PR #37, opened 2026-08-04
test: DO NOT MERGE — verify CI catches a type-invalid test fixture

kewang:
Throwaway PR verifying task 9.4 of #36. **Close without merging.**

Removes the required `vcs` field from a `WorktreeInfo` fixture in `packages/core/src/aggregate.test.ts` — the exact defect issue #16 documents.

Locally: `npm test` passes all 271 assertions, `npm run type-check` exits 2. Before #36 nothing in CI ran either check, and even after adding one, a `type-check` that skipped test files would still go green here.

Expected: **Node gates fails at the Type check step.** If it passes, the hole is not actually closed.

Refs #16

### DIFF
--- a/.agents/skills/release/SKILL.md
+++ b/.agents/skills/release/SKILL.md
@@ -1,11 +1,11 @@
 ---
 name: release
-description: Automate spek release — update CHANGELOGs, bump version, and push tag to trigger CI/CD publish. Use when the user wants to release a new version.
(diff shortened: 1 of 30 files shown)


kewang:
Verified. `Test` passed all 267 assertions and `Type check` then failed with the expected `TS2741` — the exact shape of the hole in #16, now caught. Closing without merging as intended.

PR #44, opened 2026-08-09
feat(schemas): browse workflow schemas on every surface

nthansen:
Adds a **Schemas** capability: enumerate the workflow schemas available to a repo, and render one as a diagram of its steps and their `requires` dependencies. Web, VS Code, and IntelliJ.

Implements `add-schema-browsing`, archived here with its deltas synced into `openspec/specs/`. The reasoning for anything below lives in that spec and in `design.md`.

## New public surface in `@spekjs/core`

Additive only — no existing signature changes — but the package gains surface it didn't have:

- New runtime dependency **`yaml`** (previously `cross-spawn` alone).
- New subpath export **`@spekjs/core/schema-flow`** — separate from the index because the browser bundle imports it and the index reaches for `child_process`. Being a subpath, renaming it later is breaking.
- New index exports: `listSchemas`, `readSchema`, `groupSchemaUsage`, `parseSchemaYaml`, `isSafeSchemaName`, `shortenSchemaPath`, `clearSchemaCache`, the schema types, and the CLI test seam.

## Shape

- **Core**
  - `listSchemas` / `readSchema` — CLI enumeration merged with project-local `openspec/schemas/*/schema.yaml` read from disk.
  - The CLI runner and TTL cache live in one place, `openspec-cli.ts`, shared with `schema-order.ts` — which is **−73 lines** here as a result. Previously each caller carried its own copy, and the cache's "remember failures forever" bug had to be fixed in each by hand.
- **API**
  - `GET /api/openspec/schemas` and `/schemas/:name`, plus the same pair on IntelliJ's embedded server.
  - The list aggregates two sources: project-local `openspec/schemas/` from disk, and the CLI for package and user schemas. If the CLI can't be consulted the request has still succeeded — a list was produced — so it returns 200 with what was readable plus a `degradedReason` naming the source that was missed. A 5xx would discard the project schemas that *are* available and tell the client to retry something that won't recover on its own.
  - Reading a single schema does return **404** when it can't be produced, with a `reason` separating "doesn't exist" from "couldn't look".
- **UI**
  - `/schemas` and `/schemas/:name` routes, sidebar entry, and schema badges linked throughout.
  - The workflow renders as an SVG dependency diagram rather than a card stack, so it shows which step feeds which.
- **IntelliJ**
  - Kotlin mirrors the module split (`OpenspecCli` / `SchemaCatalog` / `SchemaFlow` / `SpekCaches`) and shares the TypeScript side's test fixture.

## One call you may want to overturn

**The diagram does not draw every declared `requires`.** Where a prerequisite already reaches a step by a longer path, the direct edge is omitted (`drawableRequires`).

Given a schema declaring:

```yaml
- id: proposal
  requires: []
- id: specs
  requires: [proposal]
- id: verification
  requires: [proposal, specs]
```

the diagram draws:

```
proposal ──> specs ──> verification
```

`proposal → verification` is declared but not drawn. `proposal → specs → verification` already imposes it, so the direct arrow states nothing new while running a second line the length of the diagram. Levelling uses the **full** `requires`, so no step changes position — only the arrow is absent.

(Those declarations are `propose-spec-verify`, which has three such edges. `spec-driven` has none, so every declared edge is drawn there.)

## Testing

`type-check`, `lint`, `npm test` (396) and `./gradlew test` (177) pass locally, 0 failures.

Degradation and shadowing were verified against real conditions, not stubs: with the `openspec` binary renamed off `PATH`, `/schemas` returned 200 + `degradedReason` and still listed the project schema, while a package schema's detail returned 404 + `reason`; forking `spec-driven` into `openspec/schemas/` produced one entry sourced `project` with the shadowed package path on its detail view.

## Possible follow-up

The overview could report how many schemas are available alongside its other counts. Left out to keep this PR's scope down — it's already large — and it wouldn't be purely additive anyway: the count needs an `openspec schemas` spawn, which the spec deliberately keeps off the overview path.


kewang (changes requested):
Thanks for this — it is a lot of surface, and the reasoning in the comments and in `design.md` made it reviewable at a pace the diff size would not otherwise allow. The transitive reduction call reads right to me: keeping levelling on the full `requires` means the only thing the reduction can cost is an arrow, and the three community schemas I checked bear out that the omitted edges are the redundant kind.

Eight things I would like addressed before this goes in. The first four are cases where a designed behaviour does not hold on some host; the last four are cost and consistency.

**1. VS Code's watch refresh does not invalidate the schema cache** — `packages/vscode/src/panel.ts:155`

`notifyFileChange` only posts `{type: "fileChanged"}`. `clearSchemaCache()` appears in `handler.ts` solely inside `resync()` (line 295), which runs on a *manual* refresh. So editing `openspec/schemas/<name>/schema.yaml` in VS Code fires the watcher, the webview refetches, and `readSchema`'s 30s TTL entry is still live — the panel re-renders the pre-edit workflow while reporting a successful refresh.

The web route does this at `packages/web/server/routes/openspec.ts:71` and IntelliJ at `SpekBrowserPanel.kt:266`, both with a comment saying the refetch must not beat the invalidation. VS Code is the one host that skipped it.

**2. `MessageAdapter`'s 10s ceiling is below `getSchemas`' worst case, so the degraded-200 path is unreachable in the webview** — `packages/web/src/api/MessageAdapter.ts:74`

`request` hard-codes a 10 000 ms timeout, and `CLI_TIMEOUT_MS` is also 10 000 ms. `listSchemasUncached` awaits `openspec schemas --json` and *then* a `schema which` per schema — two serial rounds, so up to 20s. Against a slow or hung `openspec`, the extension host is still working when the webview rejects, and the Schemas page shows `Error: Request timeout: getSchemas` rather than the project-local schemas plus `degradedReason` that the whole degradation design exists to produce. Single-schema `getSchema` sits exactly on the boundary (10s vs 10s).

Whatever the fix, the webview budget needs to be derived from the CLI budget rather than coincide with it.

**3. The stretched-link overlay swallowed the tooltips already on those rows** — `packages/web/src/pages/ChangeList.tsx:58, 66, 77`

`StretchedLink`'s `after:inset-0` covers the whole card, and only `SchemaBadge` was lifted out with `relative z-10`. The `editing` badge, the `conflicts with` badge and the relative-time span all carry `title=` and all sat under the overlay now; `Dashboard.tsx` has the same pattern over `WorktreeBadge` and its date tooltip. These worked before the row became a card-wide target, so this is a regression in the jj/worktree affordances rather than anything to do with schemas. Each needs the same `relative z-10` treatment.

**4. The demo build captures the builder's machine-local schemas** — `scripts/build-demo.ts:113`

`listSchemas(REPO_DIR)` enumerates all three sources and the loop below reads every one into `schemaDetails`, absolute `path` included. `CLAUDE.md:116` records that a first build embedded twelve machine-local schemas and prescribes moving them aside before rebuilding — a manual step, on every release, for an artifact `pages.yml` uploads verbatim to the public site.

Please filter in the script (`source !== "user"`, or the repo's own plus package) so the demo payload cannot depend on who built it. A note in CLAUDE.md is the right documentation but the wrong guard.

**5. Building the catalog spawns 1 + N subprocesses, for `stageCount` alone** — `packages/core/src/schemas.ts:264`

Every summary from `parseSchemasList` has `stageCount: null`, so the `Promise.all` calls `readSchema` → `resolveSchemaPath` → `runOpenspec(["schema", "which", ...])` once per schema. With the ~12 schemas this PR's own survey mentions, a cold `/schemas` forks 13 concurrent Node processes.

It also moves the ground under the TTL rationale. `openspec-cli.ts` justifies `CACHE_TTL_MS = 30_000` as "deliberately >= the CLI timeout so an in-flight spawn can never be judged stale" — written when a compute was one spawn. Two serial rounds still fit inside 30s, so nothing is broken today, but the invariant is now "TTL >= worst-case compute" with 10s of headroom, and a third round would break it silently. Worth either restating the rationale in those terms or removing the fan-out (the CLI's list output already carries an `artifacts` array — is a level count derivable from it without the second round?).

**6. `SchemaDetail` pulls the whole catalog for one number** — `packages/web/src/pages/SchemaDetail.tsx:23`

`useSchemas()` is there only for `usage.count` in a single StatCard. The comment's premise — the catalog is already cached by the time a user arrives — holds for the in-process TTL but not for a deep link, a webview reload, a refresh after a file change (which clears the cache), or the demo and IntelliJ hosts. On a cold cache that is a full aggregated change scan plus the fan-out above, on top of the `/schemas/:name` request the page already made. A `usage` field on the definition response, or `?usage=true` on the detail route, would avoid it.

**7. Schema-cache invalidation is broader than the events that can affect it** — `packages/web/server/routes/openspec.ts:71`

The watcher admits every `.md` and `.yaml` under `openspec/`, and `clearSchemaCache()` runs *before* the 500ms debounce — so a burst of saves to a change's `tasks.md` clears the catalog once per raw event, and with the Schemas page open each debounce window re-pays finding 5. The cache is a module-level singleton too, so a watcher on repo A discards repo B's catalog. Scoping to `openspec/schemas/` and `openspec/config.yaml`, and to the watcher's own repo, would leave the behaviour intact and the cost proportionate.

**8. The Kotlin TTL cache memoises the value, not the in-flight computation** — `packages/intellij/.../core/OpenspecCli.kt:125`

`ttlCached` on the TypeScript side stores the Promise specifically "so concurrent callers arriving mid-flight share one run instead of starting a second". `TtlCache.getOrCompute` stores `V`, and the doc calls the get-miss-compute-put race benign because two callers compute the same idempotent value. That is true of the value but not of the cost: two Netty threads hitting `/api/spek/openspec/schemas` cold each run `listSchemasUncached`, i.e. 2 x (N+1) subprocesses at up to 10s each. Opening the tool window while a `spek:fileChanged` refetch is in flight is enough. Given how carefully the rest of this PR keeps the two languages stating the same rule rather than inheriting it, this one looks like an oversight rather than a decision.

---

Two smaller notes, not blocking:

- `packages/core/src/schema-flow.ts` has no test file in core's own suite. The direct coverage of `drawableRequires` / `computeArtifactLevels` / `applyStepLevel` lives in `packages/web/src/utils/schemaView.test.ts`, which imports `@spekjs/core/schema-flow` — that is `dist/`. Per CLAUDE.md's own warning, editing the rule and running `npm test` without `build:core` goes green against the previous build. This is also the one function deliberately not mirrored in Kotlin, so there is no second implementation to disagree with it. A `schema-flow.test.ts` beside `schemas.test.ts` would close both gaps.
- spek lists and renders project schemas the CLI itself rejects. I dropped a `schema.yaml` whose artifact omits the required `template` into `openspec/schemas/`: `openspec schemas` silently omits it, `loadSchema` throws `artifacts.0.template: expected string, received undefined`, and spek returns 200 and draws it. That follows from reading project schemas off disk with a lenient parser, which is the right call for a viewer that must work without the CLI — but it is worth knowing that the page can show a workflow OpenSpec would refuse to run.

---

One last thing, separate from the review, about the custom-schema side of this in particular.

I have never written a schema of my own. `spec-driven` has covered everything I do, so the project and user layers were something I knew existed and had no reason to think about — which is roughly why spek had nothing to say about schemas until now.

Testing this changed that. I wrote a throwaway schema to exercise the page, and seeing the workflow as a diagram — which steps actually gate which, which two are free to happen in either order, how much of a schema is just instruction text I would be writing myself anyway — is what made authoring one look like something worth doing rather than a configuration chore. I expect I will design one for real now.

So: thank you for the custom-schema half of this. It is the part I had the least use for going in, and the part that has changed what I plan to do with the tool.

nthansen:
All eight are addressed, plus the schema-flow test note. Specs under `openspec/specs/` are amended in place alongside the code, and the archived change's own artifacts — `tasks.md`, its delta specs, and the Impact list the release note reads — are corrected where this work overtook them, since that change is introduced by this PR rather than inherited. Gates: `type-check`, `lint`, 480 node tests, 178 Kotlin tests, 53/53 specs, 0 failures.

Two of them turned out to be wrong calls of mine rather than costs to tune — item 5, and the schema-validity note. Those and one consequence are at the bottom.

## The eight

**1 — VS Code watch invalidation.** `notifyFileChange` now discards the cache before posting `fileChanged`. Fixed together with 7, because copying the web call as it stood would have brought that finding's cost along with it.

**2 — Webview budget.** The two values now come from one definition, on a new browser-safe subpath `@spekjs/core/cli-budget`; the webview derives its ceiling from the CLI's rather than restating it. Also applied to `getChange`, which you didn't flag but sits on the same boundary — reading a change asks for its `schemaOrder`, which is an `openspec status` round.

**3 — Stretched-link overlay.** `relative z-10` on the `editing` and `conflicts with` badges, both dashboard time spans, and on `WorktreeBadge` itself, since every list rendering it is such a row. `StretchedLink`'s doc now states that the overlay intercepts hover, not only clicks.

**4 — Demo payload.** Filtered in the script, as asked — `source: "user"` plus project schemas git doesn't track, which covers the `.git/info/exclude` case too. Your parenthetical about the absolute `path` was also live; definitions now carry `displayPath` in its place. Building to a scratch path:

```
  skipping machine-local schema "behaviour-driven" (source: user)
  ... 10 more ...
  skipping untracked project schema "superpowers-bridge" (not committed)
  1 schemas (1 readable)
```

`docs/demo.html` itself is untouched.

**5 — The 1+N fan-out.** Removed, and nothing was lost with it. See below.

**6 — Catalog fetch on the detail page.** `usage` now rides on the definition response, counted from the change scan with no subprocess, and scoped to the same aggregation as the list so the two counts cannot disagree. `useSchemas()` is gone from that page.

**7 — Invalidation scope.** Narrowed three ways, none of which weaken the ordering you identified: only for paths that can change a schema read, only for the watcher's own repo, and once per debounce window instead of once per raw event.

**8 — Kotlin cache.** The entry holds a `FutureTask`, so a caller arriving mid-flight joins the run. `putIfAbsent` + `run()` rather than `computeIfAbsent`, which would hold the bin lock across a subprocess. Failures unwrap, so a shared run reports what an unshared one would.

**Note — `schema-flow` tests.** Added in core, 24 of them, including one asserting that levelling is unchanged by the reduction — the property the whole call rests on, previously only reasoned about. The transitive-reduction call itself stands unchanged; it was only ever the count that was reading the `requires` graph for the wrong reason.

**Note — schemas the CLI rejects.** Addressed, and it turned out to be the visible end of something larger — below.

## Three corrections this went on to make

**The stage count was my mistake, made in this PR and now reverted.** It invented a unit — distinct dependency levels — that has to be *derived* from the `requires` graph, and the enumeration doesn't carry `requires`:

```json
{ "name": "spec-driven", "artifacts": ["proposal", "specs", "design", "tasks"], "source": "package" }
```

Hence the `schema which` per schema you found. OpenSpec already names the thing, `artifacts:`, and counting those needs no derivation — so `/schemas` is **one CLI call for twelve schemas**, the number stays on every row, and it is `artifactCount`.

It also lines the pages up: `spec-driven` declares 4, and the Change page shows 4 for a change writing 8 files, since `discoverArtifacts` already collapses `specs/` into one artifact.

**`SchemaFlow.kt` is deleted.** Once the count stopped needing levels, nothing in Kotlin called it — no Kotlin host draws the diagram, since the tool window loads the same React SPA. Its eight tests went with it; the new core suite covers the same rules against the one remaining implementation.

**spek no longer reads schemas from disk at all.** Parsing them ourselves was my decision and it was the wrong one; your note is where it surfaced.

Deciding what a schema *is* isn't spek's to answer. A schema isn't repo content the way a spec is; it's configuration for OpenSpec's engine, resolved across three directories with precedence and shadowing between them, of which a repo holds one. I had us answering that from the one directory we can see, with a parser more forgiving than OpenSpec's — hence a `schema.yaml` it refuses, drawn as runnable. The format is versioned and these commands are still marked experimental, so the parse was also a standing bet on someone else's file layout.

Gone: the disk source, the merge, and the disk fallback in `resolveSchemaPath` — that last one could resolve a name to a *different* schema than OpenSpec would run, and said so in its own comment ("shadowing is unknown"). The list is the CLI's answer; without the CLI the catalog degrades to empty-plus-reason, exactly as `schemaOrder` already degrades rather than reimplementing OpenSpec's ordering. Parsing survives in one place only — the definition behind the diagram, which no CLI command returns — and now runs only for a schema OpenSpec has already accepted.

**One line of the description no longer holds:** with no CLI the Schemas page lists nothing, where it used to list the repo's own. I think that's the right trade — offline you'd have seen project schemas and never package or user ones, so the page was quietly claiming a completeness it couldn't have. Worth your call, since it was a deliberate design point.

One addition it forced: the Kotlin side had no CLI test seam, so its only schema coverage was of the disk scan being deleted. `SchemaCatalog.cliRunner` mirrors the TypeScript `setOpenspecRunner`, and the two new Kotlin tests drive it.

kewang:
Went through all eight against the code rather than the summary. They hold. A few notes, then one thing left.

**On the offline question you put to me: keep it as you have it. CLI-only is right.**

The fallback only ever produced anything for a repo that ships its own schemas — and that is exactly the case where it was most misleading, because it would list those and silently omit every package and user schema, with no way to know which of them shadowed what. For every other repo, spek's own included, it produced an empty list anyway. So the fallback bought a quietly incomplete answer in the rare case and nothing at all in the common one. Trading that for an honest empty plus a reason is the better deal, and it puts schemas on the same footing as `schemaOrder`, which has always degraded rather than reimplementing the CLI.

The disk fallback in `resolveSchemaPath` settles it on its own, separately from the completeness argument: resolving a name to a different schema than OpenSpec would run is a correctness bug, not an approximation. That its own comment said "shadowing is unknown" makes it a known one.

**On `stageCount`.** My finding was aimed at the cost — 1 + N subprocesses for one number. You found what was generating the cost, which is a better answer than the one I asked for: the unit had to be derived from a graph the enumeration does not carry, so the fan-out was the price of a number spek invented. `artifactCount` needs no derivation, and the enumeration already carries it. That it also makes `spec-driven` read 4 on the schema page and 4 on a change page writing eight files is the kind of agreement that only happens when the unit is the right one.

**On dropping the disk read.** Worth stating for whoever reads this later: the distinction you drew — a schema is OpenSpec's configuration, resolved across three directories of which a repo holds one, not repo content the way a spec is — is the load-bearing part, and it is why "just read the file, it is right there" was wrong. I raised it as a note about a lenient parser. You took it to where it actually came from.

**One thing left, and it is copy.**

`packages/web/src/pages/SchemaList.tsx:119` still renders "No workflow schemas were found for this repo." from `schemas.length === 0` alone. With no CLI that now prints directly under "Schemas could not be listed because the OpenSpec CLI is not available" — one line saying we could not look, the next asserting what is not there. The header count reads "0 schemas" in the same state, which is the same claim in a smaller font.

This is the distinction you kept on the detail page: `schemaUnavailableMessage` separates "does not exist" from "could not look", and the list page should not collapse it now that empty-because-degraded is a state it can actually reach. Suppressing both when `degradedReason` is set would do it — the degraded line is already saying the useful thing.

That is the last of it from me. Once that lands I will approve.

nthansen:
Fixed in `67a7ba4`.

`schemaCountClaims(count, degradedReason)` in `utils/schemaView.ts` decides the header count and the empty state together — guarding them separately at two call sites is how they came apart, since both answer the same question. As one function it is also testable, which I had written off in the change's own tasks as "a one-line render with no logic to assert without a DOM"; that note now records the correction rather than the excuse.

Verified against a dev server started with `openspec` off `PATH`: the page renders the heading, the description and the degraded line, nothing after it, in either theme. With the CLI back it reads "12 schemas" over twelve rows, and the true empty state still says so. The spec and the archived change's delta spec carry the rule and a scenario for it.

Reviewing that turned up one more of the same kind, in `6f433d9`: both servers explained their degraded 200 in a comment as "the CLI is one of two sources — project-local schemas still come off disk", the model this PR deleted. That is the reasoning that makes an empty catalog look safe to render as a fact. Comment-only, one place per host.

`useSchemas` has a single consumer, so the list is the only surface that can make this claim; VS Code and IntelliJ return the same `degradedReason` through the same SPA.

Gates: `type-check`, `lint`, 484 node tests, 178 Kotlin tests, 53/53 specs, 0 failures.

kewang (approved):
Verified `67a7ba4` and `6f433d9`. `schemaCountClaims` deciding both claims in one place is the right shape, and suppressing `unresolved` and the unresolved-default line along with them is a third instance of the same conflation I had not spotted — accusing a repo's own default schema of not existing while it sits on disk is the worst-sounding of the three.

`6f433d9` is the one I would single out. Both route comments still described the deleted two-source model, and you went after the sentence that would regenerate the bug rather than the bug. A stale comment reproducing the behaviour it describes is a failure mode worth naming.

Approving. One thing I found reading the specs, which I will fix on master under the "small fixes on merge, with credit" clause in CONTRIBUTING rather than hold the merge for it:

`openspec/specs/schema-browsing/spec.md:104-105` still says the path "SHALL be resolved through the `openspec` CLI for package schemas and directly from `<repo>/openspec/schemas/<name>/` for project-local ones", and the scenario at 118-120 has a project-local definition read "without consulting the `openspec` CLI for its path". Both are the pre-`aad2c41` model, and both contradict the same file at 237-245 ("SHALL NOT resolve the name from disk instead") and `resolveSchemaPath`, which now goes through `schema which` for every source. The archived delta carries the same pair at 98-99 and 112-114.

Worth flagging because of which direction it points: a spec saying a project schema can be located without the CLI is a specification of the resolution bug you removed. Someone making the code match the spec would put it back. Same class as the comments in `6f433d9`, one level up — and `openspec validate` passing 53/53 cannot see it, since it checks structure rather than agreement with the code.

Thanks for the two rounds. The corrections you made unprompted — the invented stage unit, and the disk read itself — were both larger than what I asked for and both right.

Issue #33, opened 2026-07-31
TaskParser.kt counts a checkbox whose line ends in a bare CR; @spekjs/core does not


kewang:
The TypeScript and Kotlin task parsers are separate implementations of one rule, and their tests
mirror each other case for case. That keeps them aligned on the cases someone thought to write down —
but not on the ones where the two regex engines disagree about what the *same pattern string* means.

## The divergence

Both parsers use the same pattern, `^- \[([ xX])] (.+)$`. Java's `$` matches before a final line
terminator; JavaScript's `$` (without the `m` flag) matches only at the absolute end of the string.
So a line ending in a bare `\r` is a checkbox to Kotlin and not to TypeScript.

Verified directly:

```
node: parseTasks("- [x] foo\r")           -> { total: 0, sections: [] }
java: Pattern.compile("^- \\[([ xX])] (.+)$").matcher("- [x] foo\r").find()  -> true
java: same pattern with \z instead of $                                      -> false
```

`SECTION_RE` (`^## (.+)$`) has the same shape and the same problem.

## When it bites

`parse()` normalises `\r\n` to `\n`, so ordinary CRLF files are fine. A line keeps a trailing `\r`
when the file uses **CR-only** line endings, or when a stray `\r` precedes a CRLF (`\r\r\n` — which
survives normalisation as `\r`). Such a file gives a different `total` in the IntelliJ plugin than in
the Web and VS Code surfaces reading the same repo.

Repos here are LF-normalised (`.gitattributes`, since beaebbb), so this needs a file that arrives
from outside that normalisation.

## Not introduced by the block-opener work

Both regexes predate it. The `BLOCK_OPENER_RE` added in a87b33d deliberately writes `\z` on the
Kotlin side where the TypeScript side writes `$`, precisely to avoid extending this, and says so in a
comment — so the new rule is aligned; these two older patterns are not.

## Suggested fix

Either strip a single trailing `\r` per line after the split, or replace `$` with `\z` in
`CHECKBOX_RE` and `SECTION_RE` on the Kotlin side. Whichever is chosen, it needs a mirrored test on
both sides, since that is the only mechanism keeping the two parsers honest.

Worth considering more broadly: a differential test that feeds generated input through both parsers
would catch this class of drift, where case-mirrored unit tests structurally cannot.

PR #50, opened 2026-08-22
feat: view non-markdown (data) artifacts in a change

nthansen:
## Summary

This PR lets spek show non-Markdown artifacts that already exist in a change.

An OpenSpec schema sets each artifact's filename through `generates:`, and not every artifact is Markdown. The `event-driven` schema declares an `asyncapi` artifact with `generates: asyncapi.yaml`, and that artifact is required. Before this change, spek discovered only root `*.md` files and the `specs/` tree, so it dropped every `.yaml`, `.yml`, or `.json` artifact. A change authored under such a schema rendered every tab except the one the schema requires.

What it does:

- The scanner discovers root-level `.yaml`, `.yml`, and `.json` files as a new `data` artifact kind. Discovery stays root-only and excludes dotfiles, so it never shows the change's own `.openspec.yaml` metadata.
- The change-detail view renders a `data` artifact as a syntax-highlighted code block, with no table of contents and no folding. The fence language comes from the file extension.
- `MarkdownRenderer` gains syntax highlighting for fenced code blocks, through `rehype-highlight`. One pipeline covers the new `data` artifacts and the code fences already inside `proposal.md` and `design.md`. Highlight colours are per-theme tokens, each checked for WCAG AA contrast in both themes.
- A `data` artifact is counted in the change-list badge and returned by full-text search, on every host. count, search, and discovery all derive from one `rootKind` classifier, so a tab shown in the view cannot be missed by count or search.

Structure: artifact discovery splits into two files. `artifact-files.ts` is the filesystem view. It lists which files are artifacts and reads no content, and it owns `countArtifacts`, `changeDirMtime`, and the search file set. `artifact-discovery.ts` reads content and builds the `ChangeArtifact` objects. The Kotlin side mirrors the pair (`ArtifactFiles.kt` + `ArtifactDiscovery.kt`), so the IntelliJ plugin stays at parity.

Notes for reviewers:

- `@spekjs/core` is additive. `ArtifactKind` gains `"data"` and `ChangeArtifact` keeps its shape, so the published type change is minor, not patch.
- A change with only Markdown artifacts is unchanged, apart from code fences now rendering highlighted.
- The discovery internals were refactored alongside the feature (a two-layer split with one `rootKind` classifier). The behaviour is unchanged and is pinned by the existing TS and Kotlin tests.
- `data` artifacts sort into their schema position under a schema that declares them (id from the filename stem, for example `asyncapi.yaml` to `asyncapi`), and trail after the declared artifacts under a schema that does not.

## Which surface(s)?

- [x] Web app
- [x] VS Code extension
- [x] IntelliJ plugin
- [ ] GitHub Action (`spekhq/spek`)
- [x] `@spekjs/core`
- [ ] `@spekjs/ui`
- [x] Docs / build / tooling

## Checklist

- [x] `npm test` passes
- [x] `npm run type-check` passes
- [x] I followed the [contributing guidelines](../CONTRIBUTING.md) and matched the surrounding code style
- [x] My editor did **not** reformat whole-file line endings (see CONTRIBUTING → Coding conventions)


kewang:
Thanks for this — the two-layer split (`artifact-files` / `artifact-discovery`) with a single `rootKind` classifier is the right shape, and mirroring it in Kotlin so count/search/discovery cannot disagree is exactly the invariant that was missing. CI is green and the feature works: I built a fixture repo with a project-level `event-driven` schema declaring `asyncapi` (`generates: asyncapi.yaml`), plus a second change under `spec-driven` carrying an undeclared `.yml` and `.json`, and confirmed the tab renders, the count includes it (5 for both changes), search reaches it (`deadLetterId` hits `payload-example.json`), and both the nested `config/settings.json` and the change's own `.openspec.yaml` are correctly excluded.

Below is everything the review turned up, most severe first. Item 1 is the one I'd call blocking — it is a claim in the PR description that does not currently hold.

---

## 1. A declared `data` artifact does not sort into its schema position

The PR description says:

> `data` artifacts sort into their schema position under a schema that declares them (id from the filename stem, for example `asyncapi.yaml` to `asyncapi`), and trail after the declared artifacts under a schema that does not.

Both halves currently produce the same result. With the sort control on **Schema order**, against a schema whose artifact list is `proposal, asyncapi, specs, design, tasks`:

```
expected:  proposal -> asyncapi -> specs -> design -> tasks
actual:    proposal -> specs -> design -> tasks -> asyncapi
```

`openspec status --change ... --json` does return the artifact:

```json
"planningArtifacts": ["proposal", "asyncapi", "specs", "design", "tasks"]
```

but `idForOutputPath` in `packages/core/src/schema-order.ts:73` strips only `.md`:

```ts
const stem = base.replace(/\.md$/i, "");
```

so `asyncapi.yaml` yields the stem `"asyncapi.yaml"`, which does not match the id discovery assigned (`"asyncapi"`, via `stripExt` in `artifact-discovery.ts:23`, which strips *any* extension). The ref is skipped, `schemaOrder` comes back with four entries instead of five, and `sortArtifacts` puts the unlisted artifact last — indistinguishable from the undeclared case.

`schema-order.ts` is not touched by this PR, so this is pre-existing code the new feature depends on rather than a regression. `packages/intellij/.../SchemaOrder.kt:132` has the identical `\.md$` strip, so IntelliJ is affected the same way.

This is the same root shape as item 6 below: filename-to-name mapping is written in three places with three different extension rules. Deriving all of them from `DATA_EXTENSIONS` would close both.

## 2. A fence with no language renders as an inline chip

`packages/web/src/components/MarkdownRenderer.tsx:289`:

```ts
const classList = className?.split(/\s+/) ?? [];
const isBlock = classList.some((c) => c.startsWith("language-")) || classList.includes("hljs");
```

react-markdown gives `<code>` no `className` for a bare triple-backtick fence with no language, and with `detect: false` rehype-highlight adds no `hljs`, so `classList` is empty and `isBlock` is false. The block renders as `<code class="bg-bg-tertiary text-code-text px-1.5 py-0.5 rounded text-sm">` — `--color-code-text` (amber in dark, `#92400e` in light) with inline pill padding.

`openspec/specs/markdown-renderer/spec.md`, as synced by this PR, states:

> A code block whose fence declares no language SHALL render as plain, uncolored code

and `design.md` D6 repeats it. `MarkdownRenderer.highlight.test.ts` only asserts the absence of `hljs` and the presence of the text, so it passes while the requirement is unmet. Detecting blockness from the `pre` parent (or handling it in the `pre` component) rather than from the class list would fix it.

## 3. File watchers still ignore `.yml` and `.json` (Web + VS Code)

`packages/web/server/routes/openspec.ts:56` and `packages/vscode/src/watcher.ts:38`:

```ts
return !filePath.endsWith(".md") && !filePath.endsWith(".yaml");
```

Editing `example.json` or `config.yml` — both shipped as fixtures by this PR — with the app open fires no SSE `/watch` event, so the tab keeps showing stale content until a manual refresh, and a newly added `.json` artifact never appears (the change-list count badge goes stale with it). `.yaml` works only incidentally: it was allowed for `openspec/config.yaml`.

## 4. IntelliJ's polling snapshot has the same filter

`packages/intellij/.../WatchPolling.kt:154`:

```kotlin
.filter { it.isFile && (it.name.endsWith(".md") || it.name.endsWith(".yaml")) }
```

`scanSnapshot` stays byte-identical when a `.json` / `.yml` artifact changes, so the tool window never repaints. Discovery, count and search all got Kotlin parity in this PR; the watcher did not.

---

## 5. `assertNever` throws during render, and there is no error boundary

`packages/web/src/pages/ChangeDetail.tsx:64` — `renderArtifact` is called eagerly for every artifact when `tabs` is built (`:241`), so one artifact with a kind the bundle does not know throws before any tab is selected. There is no `componentDidCatch` / ErrorBoundary anywhere in `packages/web/src`, so React unmounts the whole tree and the user gets a blank page instead of the previous graceful "No content".

Reachable in practice: IntelliJ's Kotlin `ChangeArtifact.kind` is a free-form `String`, its `resources/webview/` bundle is a build artifact that CLAUDE.md already documents as going stale, and `@spekjs/core`'s `ArtifactKind` is on its own published version line — it just gained `"data"`. Returning a fallback element instead of throwing keeps the compile-time guarantee (that comes from the `never` parameter alone) without the runtime cliff.

## 6. `dataLanguage` silently defaults every non-`.json` title to `yaml`

`packages/web/src/utils/dataArtifact.ts:11`:

```ts
return title.toLowerCase().endsWith(".json") ? "json" : "yaml";
```

D1/D4 argue a future artifact type is safe to add because `rootKind` classifies it and `buildArtifact`'s exhaustive switch forces a branch. That guarantee stops at `RootKind`: adding `.toml` to `DATA_EXTENSIONS` (`artifact-files.ts:11`) keeps the kind `data`, nothing fails to compile, and `dataLanguage("deploy.toml")` returns `"yaml"` — highlight.js then tokenises TOML as YAML. A total lookup keyed off the same extension list, falling back to no language rather than to `yaml`, would keep the stated guarantee true.

## 7. Kotlin `stripExt` anchors with `$`

`packages/intellij/.../ArtifactDiscovery.kt:23`. CLAUDE.md records this from #33: Java's `$` also matches before a trailing line terminator, and the schema-name allowlist already uses `\A` / `\z` for exactly this. A POSIX filename may contain a newline: for `x.yaml\n`, the JS `/\.[^.]+$/` yields `x` while Kotlin yields `x\n`, so the two hosts allocate different artifact ids for the same file and a `?tab=` link built on one misses on the other. The JS side is safe (no `m` flag) — only the Kotlin anchor needs changing.

## 8. Switching IntelliJ's index to `artifactFiles` changes which file supplies the snippet

`packages/intellij/.../SearchService.kt:61`. `searchChangesDir` matches on `slug.contains(query) || content.contains(query)` and `break`s on the first hit, so for a slug match the reported `file` / `context` come from whichever artifact file sorts first. `ArtifactFiles.artifactFiles` sorts the full set by name, so `asyncapi.yaml` now precedes `design.md` — searching a slug used to preview the design prose and now previews raw YAML. Web/VS Code push one document per file, so only the Kotlin host changes behaviour.

---

## 9. `countArtifacts` lost its first-hit short-circuit

`packages/core/src/artifact-files.ts:121`. The deleted `hasSpecsTree` carried an explicit comment that it existed for the changes-list hot path — it returned on the first `spec.md` found. The replacement calls `listSpecFiles`, which readdirs `specs/`, `existsSync`es every topic and builds a list before asking `.length > 0`. Meanwhile `scanner.ts:148` calls `countArtifacts` and the worktree election calls `changeDirMtime` for the same directory, each independently calling `rootArtifacts` and `listSpecFiles`. On ~100 archived changes across several worktrees that is four directory reads and 2N `existsSync` calls per change where the old code did fewer with an early exit.

## 10. `fencedBlock` scans the whole file on every render, including unmounted tabs

`packages/web/src/pages/ChangeDetail.tsx:47`. `tabs` maps `renderArtifact` over all artifacts, but `TabView` mounts only `activeTab.content`, so every data artifact runs a ``matchAll(/`+/g)`` plus a `replace` — a full scan and a string copy — on every re-render (tab switch, sort-mode change, watcher refresh, fold toggle). `useMemo` keyed on the artifact, or a render thunk in `tabs`, would bound it.

## 11. `rehype-highlight` registers ~37 grammars

`packages/web/package.json:28`. v7 defaults to `createLowlight(common)`, pulling java, php, ruby, swift, r, vbnet, wasm and the rest into the SPA bundle, both webview bundles, and `docs/demo.html` — which is committed verbatim (3.9 MB today) and published to Pages as-is. design.md's mitigation is "prefer a light rehype highlighter", but the subset was never narrowed. Since `detect: false` already means auto-detection never needs the other grammars, passing an explicit `languages: { json, yaml, bash, typescript, ... }` would cut most of the weight with no behaviour change.

---

## 12. D3 — worth reconsidering: a `data` title keeping its extension

Not a defect, and the decision is written down, so treat this as a discussion. D3 gives two reasons for `title: file`:

> This keeps the tab clear. It also stops a `data` tab from repeating a same-stem markdown tab's label.

The second reason is real — I verified it. With `notes.md` and `notes.yaml` in one change:

```
notes      markdown   "Notes"
notes-2    data       "notes.yaml"     <- would also be "Notes" under humanize(stripExt(file))
```

The ids dedup, the titles do not. But that requires an author to keep `foo.md` and `foo.yaml` at one change root, which is a corner; the cost is paid by every `data` tab in the common case. The tab strip currently mixes two vocabularies — `Proposal` / `Specs` / `Design` / `Tasks` are artifact names, `asyncapi.yaml` is a filename — and the schema itself calls that artifact `asyncapi` (`id: asyncapi`, `generates: asyncapi.yaml`), which is also what `openspec status` reports. CLAUDE.md already states the principle for the schema pages: `id` stays the label, because renaming it makes the view disagree with `openspec status`.

The naive fix has its own problem — `humanize("asyncapi")` gives `"Asyncapi"`, which is worse than what is there now. So something like: drop the extension in the common case and restore it only on an actual label collision, or keep the stem as the label and show the format as a small badge. Either keeps the corner case unambiguous without making it the default.

---

<details>
<summary>13-17: docs and tidiness</summary>

**13. `CLAUDE.md:409` is now inaccurate.** It states "Express only reads `.md` / `.yaml` files under `openspec/`" — the repo's stated containment property. The search route now reads everything `listChangeArtifactFiles` returns, including `.yml` and `.json`. The property itself still holds (containment under `openspec/` is unchanged); only the sentence is stale, and it is the sentence the next reviewer's judgement rests on. The PR updates the structure line and the `discoverArtifacts` bullet but leaves this one.

**14. `packages/intellij/.../Models.kt:79`** still reads `kind 為 "markdown" | "tasks" | "specs"`. The Kotlin DTO keeps `kind: String` precisely because the type cannot enforce the set, so the comment is the only statement of what the field may hold — someone adding a `when (kind)` will write three branches. `types.ts:86` was updated in the same PR.

**15. `rootArtifacts` re-implements what `listRootFiles` was extracted to state once.** `listRootFiles`'s own doc says root-only and non-dotfile are "the shared rules of every root-file listing, stated here once", but `rootArtifacts` (`artifact-files.ts:52`) open-codes `e.isFile() && !e.name.startsWith(".")` and the sort, leaving `listRootFiles` used only by the legacy `listChangeMarkdownFiles`. A future change to the shared rule has to be made twice or it applies only to the deprecated export. The trailing `.map((file) => ({ file, kind: rootKind(file.toLowerCase())! }))` also discards the kind computed moments earlier and recomputes it behind a non-null assertion; carrying `{name, kind}` through the partition removes both. Same shape at `ArtifactFiles.kt:38`.

**16. The `classList.includes("hljs")` fallback is unreachable, and its comment is inaccurate.** rehype-highlight only adds `hljs` to a node it already read a `language-*` class from (or auto-detected, which `detect: false` disables), so the "highlighted block whose language class the highlighter dropped" case cannot occur. The comment also says the old `startsWith` path made token spans "render as `[object Object]`" — the inline branch renders `{children}`, not `text`, so the spans rendered correctly; only the `specTopics.includes(text)` lookup saw the stringified array.

**17. `packages/web/server/routes/openspec.search.test.ts:33`** calls `fs.mkdtempSync` per test with no `t.after` / `rmSync`, so each `npm test` leaves two populated `spek-search-test-*` trees behind. The existing `after()` closes only the HTTP server, so the asymmetry reads as intentional. The Kotlin mirror (`ArtifactDiscoveryTest`) does clean up via `@AfterTest`.

</details>

---

Happy to share the fixture repo if it helps — it is a small OpenSpec repo with a project schema at `openspec/schemas/event-driven/schema.yaml` declaring `asyncapi` (`generates: asyncapi.yaml`), one change using it, and a second under `spec-driven` carrying an undeclared `retry-policy.yml` and `payload-example.json`. Items 1, 2 and 12 reproduce on it directly.

nthansen:
All 17 items addressed in `c52783b`, and I ran the changes through an adversarial review pass that surfaced a few more (noted at the end). Rundown:

**Blocking / correctness**
- **1** — Fixed. `idForOutputPath` now strips the last extension (mirroring `discoverArtifacts`' `stripExt`), so a declared `data` artifact resolves to its id. Same fix in `SchemaOrder.kt`. The adversarial pass caught a follow-on: when a markdown sibling shares the stem (`asyncapi.md` + `asyncapi.yaml`), the outputPath mapped to the markdown id — now resolved to the data id via a data-filename→id map (a `data` artifact's title is its filename). Both hosts, with tests.
- **2** — Fixed. Detecting blockness from the source line-span was my first cut, but it wrongly promotes multi-line *inline* code — so it keys off the trailing newline `mdast-util-to-hast` appends to code blocks. Also handles the empty-fence case.
- **3 / 4** — Fixed. All three watchers now react to `.yml`/`.json`, derived from one shared `DATA_EXTENSIONS` (no drift), matched case-insensitively to align with discovery.
- **5** — Fixed. `renderArtifact` returns a fallback element instead of throwing; the compile-time exhaustiveness (`never`) is kept.

**Item 6 / 7 / 8 / 9 / 10**
- **6** — Total lookup keyed off the extension; unknown → no language (never `yaml`). Data tabs now show a format badge instead of the extension in the label (addresses **12**).
- **7** — Kotlin `stripExt` anchored with `\z`. (For this greedy `[^.]+` pattern `$` and `\z` actually coincide, since the class eats a trailing newline — kept `\z` for the repo convention.)
- **8** — Fixed. Search iterates markdown-first, so a slug match previews prose again; data stays reachable for content-only matches. Added a `SearchServiceTest` (this path had no coverage).
- **9** — `hasSpecsTree` regains its first-hit short-circuit (both hosts).
- **10** — Only the active tab renders, bounding the `fencedBlock` scan.

**Item 11 — the suggested fix doesn't shrink the bundle**
Passing `languages: {...}` to `rehype-highlight` measured *identical* bundle size, because `rehype-highlight` statically imports lowlight's `common` (37 grammars) as its fallback — rollup can't drop it regardless of the option. To actually cut it I replaced `rehype-highlight` with a ~30-line plugin driving `createLowlight` directly (pulls only `highlight.js/lib/core`), registering the grammars this repo uses. Same `hljs`/`--color-hl-*` output, so no CSS or contrast-test change. Result: **~68% off the highlighting footprint** (179→~50 KB raw / 54→~17 KB gzip), across the SPA, both webview bundles, and `docs/demo.html`.

**13–17** — All fixed: `CLAUDE.md` containment sentence, `Models.kt` comment, `rootArtifacts` kind-carry, the unreachable `hljs` branch + comment, and search-test temp-dir cleanup.

Additional issues the adversarial pass found and fixed: the `typescript` grammar doesn't alias `js`/`jsx` (added `javascript`); empty fence misclassification; case-sensitive watchers; and the missing search-snippet test.

kewang (approved):
All 17 addressed in `c52783b`. I re-verified the ones I could exercise directly, against the same fixture repo:

- **1** — `schemaOrder` now returns `["proposal","asyncapi","specs","design","tasks"]` and the tab strip under Schema order matches; the undeclared `.yml`/`.json` in the other change still trail. Your follow-on holds too: with an `asyncapi.md` sibling present, the outputPath resolves to the data id (`asyncapi-2`) and the markdown one correctly trails as undeclared.
- **2** — the language-less fence renders plain now. Rejecting the source-span approach was right; multi-line inline code would have been misclassified.
- **3 / 4** — live SSE test on the Web watcher: `.json`, `.yml`, `.yaml`, `.md` each fire, `.txt` does not, and an uppercase `UPPER.JSON` fires too — the `toLowerCase()` alignment with discovery was a good catch that I hadn't asked for. Kotlin tests green. I also checked `SpekBrowserPanel`'s other watch path: `setupWatchServiceWatcher` filters nothing, so the gap really was polling-only, and that is what you fixed. No other extension filter is left outside the classifier.
- **12** — stem plus format badge, extension restored only on collision, is exactly the shape worth having. `dataTabNames` being pure and unit-tested is better than what I suggested.

Two corrections on my side:

- **11 — you were right and I was wrong.** `rehype-highlight@7.0.2` does `import {common, createLowlight} from 'lowlight'` unconditionally, so `languages` could never have dropped those grammars. I measured your replacement across the whole web bundle: **799 KB → 677 KB raw, 248 KB → 212 KB gzip**, which lines up with the highlighting-footprint numbers you quoted. Worth the extra ~30 lines.
- **7 was a false positive from me.** For the greedy `\.[^.]+$` pattern, `[^.]` already consumes the trailing terminator, so Java's `$` and `\z` land in the same place — I checked `\n`, `\r\n`, an embedded newline and U+0085 on a real JVM and they agree in every case. The CLAUDE.md rule is real but does not bite here. Keeping `\z` for consistency is fine.

I did not independently exercise 5, 8, 9, 10 and 13–17 — each has a test and CI is green. VS Code's watcher I confirmed by reading rather than running: the predicate is character-identical to the Web one and imports the same constant.

Thanks for the thorough turnaround, and for pushing back on 11 with a measurement instead of taking the suggestion at face value. Approving.

PR #55, opened 2026-09-22
feat(diagrams): render mermaid in markdown and as a .mmd artifact

gokul-kulkarni:
## Summary

Renders Mermaid in the viewer — `docs/feature-ideas.md` #2.

Two entry points, one implementation: a ` ```mermaid ` fence in any Markdown artifact, and a root
`.mmd` / `.mermaid` file as an artifact of its own (a tab, counted in the change's badge, and in the
search corpus).

**The decision worth your attention: only the Web build draws.** The VS Code, IntelliJ and demo
builds are single-file IIFE bundles with `manualChunks: undefined`, so a dynamic import of Mermaid is
*inlined* rather than deferred. Measured:

| bundle | before this PR | **this PR** | if Mermaid were inlined (rejected) |
|---|---|---|---|
| VS Code / IntelliJ webview | 718,653 B | **719,158 B** (+505) | 5,952,548 B (+5.23 MB) |
| `docs/demo.html` | 4,695,310 B | **≈ unchanged** (+505 B when next rebuilt) | ≈ 9.9 MB *(estimated)* |
| Web app entry | 688,793 B | **689,047 B**, Mermaid in lazy chunks | — |

The third column is the counterfactual that drove the decision, **not what this PR does**. Every
figure is measured except the demo estimate, which is extrapolated — the demo is only rebuilt at
release time, and `pages.yml` publishes what is committed, so this PR does not touch it.

Inlining would cost +5.23 MB per single-file surface, on a `docs/demo.html` that is committed to the
repo on every release. Trimming does not rescue it cleanly — the four heavyweights (`elk` 1.47 MB, `cynefin` 0.70,
`cytoscape` 0.44, `katex` 0.26) are only 2.87 MB of it, and going further means deciding which diagram
types a VS Code reader is allowed to see.

So those three exclude Mermaid outright and show diagram **source**, specified as how that surface
shows a diagram rather than as a failure. Two mechanisms, doing two jobs: `define:
{ __SPEK_DRAWS_DIAGRAMS__ }` decides what the reader sees, `resolve.alias` to an `export default null`
stand-in decides whether the bytes exist (tree-shaking a dynamic import behind a flag is a hope; an
alias is a fact). Measured after: 719,158 B, **+505**. `diagramBuilds.test.ts` guards both, because a
bundle eight times too big fails no type-check, no lint and no other test — it just lands in `docs/`.

**This is a product call, not a technical one.** If you'd rather every surface drew, or drew a trimmed
set, say so and I'll redo it — the specs and the build configs are the only things that move.

### Other notes

- **`@spekjs/core`**: `diagram` joins `RootKind` beside `data` — *"show this file's text"* vs *"show
  what this file describes"*. `ChangeArtifact`'s shape is unchanged, so this is additive: **minor, not
  patch**. The Kotlin mirror is in the same PR.
- **Palette**: Mermaid writes our colours into an SVG that does not exist until a reader opens the
  page, so `contrast.test.ts` structurally cannot see them — it would report nothing, which is
  indistinguishable from reporting nothing wrong. `utils/diagramTheme.ts` is therefore the declared
  set of colours handed over, each naming its token and floor, and the check measures *that*. This is
  the SVG-attribute gap `theme-toggle` already records, arriving by a second route; the spec delta
  states the general rule.
- **Security**: `securityLevel: "strict"`, and `bindFunctions` is deliberately never called.
- **Folding**: drawing is deferred to `IntersectionObserver`, since Mermaid lays out by measuring the
  DOM and `spec-section-folding` renders every scenario inside a closed `<details>`.
- **Tests**: the state machine and zoom maths are extracted to `utils/` and tested directly. The web
  tests are `node:test` + `renderToStaticMarkup`, which runs no effects, so a stubbed Mermaid module
  would never be called — this follows the `refreshTracker` / `foldSections` precedent.

### Process

CONTRIBUTING says to open an issue first for something this size — I went straight to a PR, so if
you'd prefer to agree the approach before reviewing 3.8k lines, I'm happy to close this and open an
issue instead. **Equally happy to split it**: the `.mmd` artifact kind (core + Kotlin + search) is
independent of the fence rendering, and either half stands alone.

### Verification

Exercised in the running Web app: both themes, the failure path (reason + source, page intact), a
diagram inside a folded scenario, an oversized diagram not widening the page, the zoom/pan and source
controls, and drawn colours matching the declared tokens exactly. No console errors or CSP violations.

Verified in the **VS Code** Extension Development Host: the webview shows diagram source, as that
build does not draw.

**Not yet verified: the IntelliJ tool window.** The built webview bundle contains no `dagre` /
`cytoscape` / `katex` / `sequenceDiagram` and the observer effect dead-code-eliminates to
`useEffect(()=>{},[])`, so the exclusion holds there too, but the host has not been run.

## Related issue

None.

## Which surface(s)?

- [x] Web app
- [x] VS Code extension
- [x] IntelliJ plugin
- [ ] GitHub Action (`spekhq/spek`)
- [x] `@spekjs/core`
- [ ] `@spekjs/ui`
- [x] Docs / build / tooling

## Checklist

- [x] `npm test` passes (core 448, ui 33, web 324; plus `./gradlew test` 244, JDK 17)
- [x] `npm run type-check` passes
- [x] I followed the [contributing guidelines](../CONTRIBUTING.md) and matched the surrounding code style
- [x] My editor did **not** reformat whole-file line endings


kewang (changes requested):
Thanks for this — the write-up made the trade-offs easy to follow, and the build guard, the declared palette and the extracted state machine are exactly the kind of thing this repo values. I merged current `master` into the branch locally: no conflicts, and `build` / `type-check` / `lint` / `npm test` / `./gradlew test` / `openspec validate --specs` all pass. I'll approve the CI runs once you push.

## The product call: VS Code and IntelliJ should draw too

You asked, so: I'd like every surface that *can* draw to draw. The demo can stay source-only.

The single-file constraint comes from the build config (`format: "iife"`), not from the hosts:

- **VS Code**: a webview can load any file under `localResourceRoots` through `asWebviewUri`, and `panel.ts` already does this for the entry script. If the webview build emits ESM, is loaded with `<script type="module">`, and `webview.cspSource` is added to `script-src` next to the nonce, then `import("mermaid")` becomes a lazy chunk shipped inside the vsix and fetched only when a document holds a diagram.
- **IntelliJ**: the built-in server already serves everything under `/spek/webview/`, so extra chunks need no new route.
- **Demo**: this one really is a single file, and it is committed on every release, so it keeps showing source as it does now.

Since both entry points (the fence and the `.mmd` tab) go through `MermaidDiagram`, this should be a build/host change rather than a per-mode one. What I'd expect to move:

1. `vite.webview.config.ts` / `vite.intellij.config.ts`: ESM output with chunking, `__SPEK_DRAWS_DIAGRAMS__: "true"`, no alias.
2. `panel.ts`: module script plus the CSP change. Check that `vsce package` picks up the chunks.
3. `diagramBuilds.test.ts`: the thing to guard becomes "Mermaid is not in the entry bundle" rather than "this build does not draw". The size risk you identified is still real; only the mechanism changes.
4. `diagram-rendering` spec (main spec and the archived copy) and the CLAUDE.md paragraph.
5. Verify in both hosts. The CSP is where I'd expect surprises: Mermaid puts a `<style>` into the SVG, and KaTeX-based diagrams load fonts, so `font-src` may need a look.

## Bugs to fix before merge

1. **A failed diagram leaves Mermaid's error graphic on the page** (`MermaidDiagram.tsx` ~L137). On a parse error Mermaid 12 renders its "Syntax error" bomb into a temp `div` appended to `body`, and throws before `removeTempElements()`. The cleanup effect runs only on unmount, so the graphic stays at the bottom of the page. `suppressErrorRendering: true` in `initialize` avoids it.
2. **Ctrl/Cmd+wheel zooms the whole page too** (~L178). React registers `onWheel` as a passive listener (since React 17), so `preventDefault()` is ignored. It needs a native `addEventListener("wheel", h, { passive: false })` on a ref.
3. **Fenced diagrams remount on every re-render** (`MarkdownRenderer.tsx` ~L441). `components` is an object literal in JSX, so the `div` override is a new component type each render. Any watcher refresh or sort change unmounts every fenced `MermaidDiagram`, discarding the drawing, zoom/pan and source toggle, and draws again. (`.mmd` tabs are unaffected.) Hoisting the override to module scope fixes it.
4. **Schema-order sort drops a diagram that shares a stem with a Markdown file** (`packages/core/src/scanner.ts:291`, and `ChangeReader.kt:44`). `dataFileToId` maps only `kind === "data"`. With `flow.md` + `flow.mmd` and a schema that generates `flow.mmd`, the output path falls through to the stem and resolves to the Markdown `flow`. Reproduced with the built core: `resolveSchemaOrder` returns `['flow']`, and `['flow', 'flow-2']` once diagrams are mapped. This is the collision case the PR sets out to handle; the other "keeps its extension" sites were updated, this one wasn't.

## Should fix

5. **Palette gaps the declared table doesn't catch** (`diagramTheme.ts`):
   - Sequence `autonumber` draws `sequenceNumberColor` (text-primary) on a `signalColor` (text-secondary) circle: about 2.1:1 dark and 2.4:1 light.
   - The "no silent defaults" rule isn't enforced. The base theme still supplies `doneTaskBkgColor: "lightgrey"`, `critBkgColor: "red"`, `altSectionBkgColor: "white"` and others. A dark gantt draws light text on lightgrey (~1.1:1), and `pie2` / `pie3` resolve to the same colour.
   - `diagramTheme.test.ts` only checks keys that appear in both lists. It never checks that every base variable is in one of them.
6. **`invalidated` moves an `idle` diagram straight to `drawing`** (`diagramState.ts` ~L86). A theme toggle or an edit starts a draw, and the chunk load, for every diagram the reader never scrolled to, including ones in closed scenarios. Under `<StrictMode>` in dev, the second effect run dispatches `invalidated` on mount, so everything draws immediately. Leaving `idle` unchanged on invalidation fixes it, and a test for that case would help.
7. **The closed-`<details>` premise doesn't hold.** `mermaid.render(id, text)` without a container lays out in a temp `div` under `document.body`, not inside the `<details>`, so no measurement is zeroed. The observer is still worth keeping for laziness. Please state that as the reason in the component and in CLAUDE.md, and disconnect the observer after the first `shown`.
8. **A rejected `import("mermaid")` is memoised for the page's lifetime** (~L75). One transient chunk failure, such as a server rebuilt while the tab is open, means no diagram draws until a full reload. Please drop the promise on rejection. CLAUDE.md records the same "remember failures forever" bug in the CLI cache.
9. **Zoom-at-cursor drifts** (~L182). The origin is measured from the padded outer box (`p-4`), but the transform applies to the inner div, so the point under the cursor moves ~16·(1−1/r)/s px per step.
10. **The Source toggle is disabled rather than omitted** where there is nothing to switch to (failed diagrams, and the demo). It has no disabled styling and still takes the hover accent, which contradicts "No control leads nowhere".

## Nits

- A theme toggle during an in-flight draw starts two full renders: the draw effect depends on `theme` / `source` as well as `generation`. Keying it on `generation` alone would drop one.
- `MarkdownRenderer.tsx` ~L314: the new comment is in Chinese. Per CLAUDE.md, new comments are in English, even where they extend an existing Chinese list. The list is also introduced as having two reasons.
- CLAUDE.md's Security bullet still says Express reads only `.md` / `.yaml` / `.yml` / `.json`; `.mmd` / `.mermaid` belong there now.

## Not for this PR

`mermaid` as a runtime dependency means the GitHub Action's `npm ci` installs ~150 MB (mermaid alone is 124 MB) for a build that aliases it away. `action.yml` has some sharp edges, so I'll take that on separately. No change needed here.

Please don't touch CHANGELOGs or versions; those are done at release. The branch is 9 commits behind `master` and branch protection requires it to be up to date, so please merge or rebase when you push.

kewang:
Thanks — that's a thorough round, and every item from the review is addressed. CI is green. We ran both hosts for real: IntelliJ looks good, but VS Code has one blocking issue that the browser-served check couldn't show, plus two palette gaps that only show up once the charts are drawn.

## Blocking: the first diagram on a page fails in VS Code

In the Extension Development Host (VS Code 1.134, both themes), the first diagram to draw on a page shows:

> This diagram could not be drawn: Unable to preload CSS for /assets/index-pHJ-Qjoe.css

Diagrams after it draw fine. The cause is Vite's preload helper in the webview entry, which builds dependency URLs from an absolute base (`return "/"+e`). In the webview, `/assets/…` resolves against the `vscode-webview://` origin, where nothing is served, so the CSS preload errors and the import rejects. Later diagrams skip the preload because the `<link>` is already in the document. Serving the build from an origin root, as the earlier check did, makes `/assets/` resolve, which is why it passed there.

A relative base should fix it: `base: "./"` in `vite.webview.config.ts`, so the helper resolves dependencies against `import.meta.url`. Then `panel.ts`'s rewrite needs to match `./assets/…`. IntelliJ's absolute `/spek/webview/` base is correct for its own server, and it drew fine there.

## Palette: drawn colours still fail in gantt and pie

- **Gantt, critical task**: the label measures ~1.7:1 in light (dark text on `#c10007`) and ~2.3:1 in dark (light text on `#ff6467`). The text-on-fill pair isn't measured.
- **Gantt, active task (light)**: a white bar on the near-white chart background is barely visible, and it is the only mark of that state.
- **Pie**: slices and legend swatches can't be told apart in either theme. They are near-black shades in dark and near-white in light, because `pie1…` derive from the surface tokens. The legend is the only key, so the chart can't be read. If a categorical ramp is out of scope here, I'd rather see pie (and any other category-coloured diagram) declared as a known limitation in `DECLARED_DEFAULTS` / the spec than shipped looking correct and unreadable. Your call on which.

## Minor

`diagramBuilds.test.ts`'s "Mermaid is not in the webview entry bundle" test skips in CI. `ci.yml` runs `npm test` before any build and never runs `build:webview`, so it always takes the skip branch, and its comment ("CI and anyone who has run a build get the check") doesn't hold. Either add a `build:webview` step before the tests in the `Node gates` job (a new step, not a renamed job, since the job name is a required check), or correct the comment. It also covers only the VS Code entry, not IntelliJ's.

